PR #2, opened 2019-05-07
Adding travis config using per-project env

jsignell:
This PR :
- Adds a travis file that sets up the mechanism for having a job for each project.
- Improves the attractors example by pinning and removing comments.
- Adds bay_trimesh as an example of using a datafile (not using intake).
- Starts a README with the "running locally" info.


jsignell:
@jbednar @philippjfr @jlstevens I wanted to solicit feedback on this work so far and see if it seems like a reasonable path to take.

README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+Then start a jupyter notebook as usual:
+
+```bash
+jupyter notebook
+```
+
+**NOTE:** If the notebook depends on data files, you will need to
+download them. The URLs are defined in anaconda-project.yml
```
jbednar:
```suggestion
download them explicitly if you don't use anaconda-project, by
extracting the URLs defined in anaconda-project.yml and saving 
the file(s) to this directory.
```

attractors/anaconda-project.yml
```diff
@@ -1,50 +1,39 @@
-# This is an Anaconda project file.
-#
-# Here you can describe your project and how to run it.
-# Use `anaconda-project run` to run the project.
-#
-# It also works as an environment file.
-# Use `conda env create --file anaconda-project.yml` to create env
 name: attractors
```
jbednar:
```suggestion
# After installing anaconda-project, this project can be reproduced using:
#    anaconda-project run
name: attractors
```

bay_trimesh/anaconda-project.yml
```diff
@@ -0,0 +1,53 @@
...
+downloads:
+  DATAFILE:
+    url: https://s3.amazonaws.com/datashader-data/Chesapeake_and_Delaware_Bays.zip
+    filename: data
+    unzip: true
+
+env_specs:
+  pyviz-topics-bay-trimesh:
+    channels: []
+    packages: []
```
jlstevens: @jsignell I have stumbled upon this file as a good example for how to use anaconda-project.yml for topics. This might have already been resolved as a question (I've not been following any discussion here) but do you need `channels: []` and `packages: []` here? Haven't they been declared at the top level?

bay_trimesh/anaconda-project.yml
```diff
@@ -0,0 +1,53 @@
...
+  test:
+    channels:
+    - pyviz
```
jlstevens: Similarly, is this declaration needed here given that the following is declared at the top of the file?

```
channels:
- pyviz
```

bay_trimesh/anaconda-project.yml
jsignell: I will do a test.

jsignell: works :)

jbednar: You say it works, but those lines still seem to be here?  Same for the pyviz channel spec below?

bay_trimesh/anaconda-project.yml
```diff
@@ -0,0 +1,53 @@
...
+
+platforms:
+- linux-64
+- osx-64
+- win-64
+
+commands:
+  notebook:
+    notebook: bay_trimesh.ipynb
+    env_spec: pyviz-topics-bay-trimesh
```
jbednar: Is the env_spec required here?  I'd like this file to be pared down to the absolute essentials, because otherwise unneeded stuff will just get copied everywhere.

jsignell: right. I just pushed.

bay_trimesh/anaconda-project.yml
```diff
@@ -0,0 +1,53 @@
...
+  test:
+    unix: pytest --nbsmoke-run -k *.ipynb --ignore envs
+    windows: pytest --nbsmoke-run -k *.ipynb --ignore envs
+    env_spec: test
```
jbednar: It's important to be able to specify commands per platform, but given they are the same here and will generally be the same, is there no way to specify the command just once?  It's also ambiguous what will happen on osx; is that considered unix?  If we do need to have both specs, please align them so that we can see at a glance that the commands are identical.

jsignell: I can use a reference if they are both needed. Let me check

jbednar: In this case with a single-line command, a reference is probably more confusing than having it copied, as long as the two copies are lined up and visually identical.  But if there's a way to specify it for all platforms as a default that can be overridden per platform, that's the way I'd rather do it in general.

bay_trimesh/anaconda-project.yml
jbednar: What was the resolution here?

bay_trimesh/anaconda-project.yml
jsignell: Ok so osx counts as unix and I believe you need to specify a command per platform. So I will align them visually.

jsignell: I am imagining that there will be some special anaconda-project init for these example projects and they will autopopulate this section.

bay_trimesh/anaconda-project.yml
jsignell: I misread. I'll check now.

bay_trimesh/anaconda-project.yml
jbednar: Ok, thanks.  Please normalize the spacing before "&deps" and "*deps" across the files; seems like it varies.  Probably just one space there is fine; it's the least surprising.

jsignell:
I decided to leave the tests in the project level anaconda-project.ymls. It was getting very verbose and brittle to move a general version up to the top level and I'd rather have each project be responsible for itself.

jbednar:
Ok.

jsignell:
Ok I have verified that these work on AE. Would it be ok to merge this and carry on work with the top level testing and website creation?

jbednar:
Please do merge this and reorganize the other files to match this structure, so that we can then get the top-level setup in place.

PR #7, opened 2019-05-15
Adding census dir with gerrymandering and census in it

jsignell:
Not sure how people will feel about this one. Basically it is a small grouping of gerrymandering and census since they use the same dataset. I think it makes sense in terms of the concept of topics. 

I also added a mechanism for skipping tests for certain examples. This is useful in cases like this where the data download is so massive that if isn't feasible. This is a a stop-gap until I figure out a new way to implement small datasets for testing (this might only happen in top level testing).


jbednar:
Grouping census and gerrymandering together makes sense for the datasets, but gerrymandering requires GeoViews, while census only uses tiles and thus should be updated to use HoloViews only.  Given that the environment takes a lot more time and is much moire trouble to set up when GeoViews is installed, does it still make sense to combine them?  E.g. can gerrymandering easily add GeoViews to what the environment here otherwise includes?

Seems like now is when we need to figure out the plan for testing to use small datasets.

jsignell:
Yes they could have separate environments with gerrymandering being a superset of census. Not quite sure how that would work for testing, but I can check it out.

jsignell:
Ok so this comes back to the idea of grouping examples or keeping them separate. I am still thinking about this in terms of "topics" which implies that a project is about a particular thing. In this case, both these notebooks are about similar things (using the same dataset is a proxy for being about the same thing). So by grouping them we make it easy for someone to find out about the other notebook once they've found the first and we allow them to depend on the same dataset.

BUT they have slightly different dependencies, they even have slightly different data dependencies. The env thing could be resolved by creating another env-spec that is a superset of the first, but then what about testing? And which do we want to thumbnail some day? 

There are several ways to resolve this:
 1. Group them, but only do one env and accept that the env that we build to run census will not be the absolute minimum required.
 2. Group them and do separate envs, with separate envs for testing as well (this will get long and not DRY in the yaml)
 3. Keep them separate and accept that we will have to download the data separately in each project.
 4. Keep them separate and try to hack together some central data storage. The benefit of this is that we could use that central data storage as a place to store miniature datasets for testing. The con is that if a user has just an example dir, the data will not be installed in that dir (It will be up a level). Also I don't know how this data dir would work on AE (it would probably have to get changed before uploading).

jsignell:
Went with option 3

PR #8, opened 2019-05-16
Separate environments but still grouped

jsignell:
Option 2:
> Group them and do separate envs, with separate envs for testing as well (this will get long and not DRY in the yaml) also all the data downloads will happen on first run even if they are only required for one of the notebooks.


jsignell:
Went with #10 instead

PR #9, opened 2019-05-16
Creating a mechanism for using stubs rather than real data when testing

jsignell:
Haven't added any data to the top level data dir, so tests should fail.

### DIFF
--- a/bay_trimesh/anaconda-project.yml
+++ b/bay_trimesh/anaconda-project.yml
@@ -36,7 +36,7 @@ commands:
 
 variables: {}
 downloads:
-  DATAFILE:
+  DATA_DIR:
     url: https://s3.amazonaws.com/datashader-data/Chesapeake_and_Delaware_Bays.zip
     filename: data
     unzip: true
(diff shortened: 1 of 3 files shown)


bay_trimesh/anaconda-project.yml
```diff
@@ -36,7 +36,7 @@ commands:
 
 variables: {}
 downloads:
-  DATAFILE:
+  DATA_DIR:
```
jbednar: Just DATA?

jsignell:
Another downside of the approach in this PR is the env vars are per dataset, so there would need to be an override for each dataset not necessarily just one for the whole data_dir. 

A downside of the other approach is that there is a risk of overwriting the real data with test data. So this is something that we would only ever want to do on travis and we'd need checks to ensure that the real data isn't there before doing a copy.

jsignell:
>  what if we use Travis to put a different file in the same location the notebook is already expecting it?

See https://github.com/pyviz-topics/examples/pull/11

jsignell:
Closing in favor of #11 which in addition to being less visible to users, also can be extended to support intake catalogs.

PR #11, opened 2019-05-20
Copy the test data into the example dir before running tests

jsignell:
This is the other option mentioned in https://github.com/pyviz-topics/examples/pull/9#pullrequestreview-239505941


jsignell:
I think this is the right approach. I'll extend it to add some data.

jsignell:
I guess one remaining question is if this "small" data should be included in the github repo or put up on s3? Maybe s3 is better since most people won't want it, but it does make it slightly trickier to add an example.

dodo.py
```diff
@@ -13,3 +15,73 @@ def task_ecosystem_setup():
...
+
+def task_small_data_setup():
+    """Copy small versions of the data from test_data"""
+
+    def copy_test_data(root='', name='attractors', test_data='test_data'):
+        print('Copying test data for {} ...'.format(name))
+        paths = _prepare_paths(root, name, test_data)
+
+        if not os.path.exists(paths['test']):
+            print("Fail: No test_data found for {} in {}".format(name, paths['test']))
```
jbednar: How will these failing conditions be detected by the tests, if they are just printing "Fail"?

jsignell: They won't. This was for debugging purposes. I should be raising proper errors.

jsignell:
This is working how I want now. Just need to make a small version of the bay_trimesh data and then I'll merge

nyc_taxi/anaconda-project.yml
```diff
@@ -0,0 +1,48 @@
+# After installing anaconda-project, this project can be reproduced using:
+#    anaconda-project run
+name: nyc_taxi
```
jbednar: Maybe the initial comment in these files could be simpler:

```
# To reproduce: install 'anaconda-project', then do 'anaconda-project run'.
```

jbednar: Or, if it has to fit into 80 characters exactly,

```
# To reproduce: install 'anaconda-project', then 'anaconda-project run'
```

jsignell:
In the interest of migrating all the projects quickly I'm going to merge this and add the small data in a follow on PR.

PR #13, opened 2019-05-21
Moving landsat

jsignell:
Except for 5a99315 this is a minimal example of making an example out of a notebook including adding small test data.


jsignell:
Just so I don't lose track. The notebook that I used to make the small data is [here](https://gist.github.com/jsignell/996bef9da8c2c5538df1f30aa4de0074)

PR #20, opened 2019-05-23
Intake-ifying osm

jsignell:
This PR:
 - moves osm-1billion
 - makes that the default notebook for the project
 - intake-ifies that notebook to allow download only when called
 - adds the infrastructure for intake test catalogs

Blocked on https://github.com/intake/intake-parquet/pull/11 and intake-parquet release.

### DIFF
--- a/osm/anaconda-project.yml
+++ b/osm/anaconda-project.yml
@@ -29,12 +29,12 @@ commands:
     description: Datashading 2.7-billion-point Open Street Map database
     notebook: osm.ipynb
   test:
-    unix:    pytest --nbsmoke-run -k *.ipynb --ignore envs
-    windows: pytest --nbsmoke-run -k *.ipynb --ignore envs
+    unix:    pytest --nbsmoke-run --ignore envs
+    windows: pytest --nbsmoke-run --ignore envs
     env_spec: test
   lint:
-    unix:    pytest --nbsmoke-lint -k *.ipynb --ignore envs
-    windows: pytest --nbsmoke-lint -k *.ipynb --ignore envs
+    unix:    pytest --nbsmoke-lint --ignore envs
+    windows: pytest --nbsmoke-lint --ignore envs
     env_spec: test
 
 variables:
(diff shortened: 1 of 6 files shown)


osm/anaconda-project.yml
```diff
@@ -50,10 +50,11 @@ env_specs:
     packages:
       - holoviews
       - intake
-      - intake-parquet
       - ipywidgets
       - jupyter
       - tqdm
+      - pip:
+        - git+https://github.com/intake/intake-parquet.git@jsignell/spatial#egg=intake_parquet
```
jbednar: intake-parquet isn't on conda?

osm/osm-1billion.ipynb
```diff
@@ -39,8 +39,25 @@
...
+    "cat = intake.open_catalog('./catalog.yml')\n",
+    "data_entry = cat.osm_one_billion\n",
+    "data_entry"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that the file isn't downloaded yet. The following step will take some time:"
```
jbednar: How do you know at this point whether it's been downloaded yet? Won't it be cached?  Seems like it should say instead that "Note that the first time this cell is executed, the file will take some time to download, but subsequent runs will skip that step".

test_data/osm/catalog.yml
```diff
@@ -0,0 +1,6 @@
+sources:
+  osm_one_billion:
+    description: Test data points to same fake osm-3billion file
```
jbednar: Same as what?  Try to reword in a way that makes sense when reading just this one file.

osm/anaconda-project.yml
jsignell: This points to a branch. The spatial isn't merged or released yet.

jsignell:
There is a lingering issue here where the urlpath that is returned in https://github.com/intake/intake-parquet/blob/f029a36ba5c7a644b4faebd2a40d6ec21dbc5681/intake_parquet/source.py#L134 is actually a list of paths where the first one is the path to the parquet dir which is all that you actually need. This crops up when you run `df.persist()`

jsignell:
Closing this for now, although I won't delete the branch yet.

PR #26, opened 2019-05-29
Moving opensky

jsignell:
- [x] Put in dir with matching name
- [x] Added anaconda-project
- [x] Added env info and data downloads
- [x] Ran locally
- [x] Moved data preparation scripts into own files
- [x] Added test data
- [x] Added opensky to travis.yml


.travis.yml
```diff
@@ -33,6 +33,8 @@ jobs:
       if: (commit_message !~ /^.*(only:).*$/) OR (commit_message =~ concat(^.*,env(DIR),.*$))
     - env: DIR=nyc_taxi
       if: (commit_message !~ /^.*(only:).*$/) OR (commit_message =~ concat(^.*,env(DIR),.*$))
+    - env: DIR=opensky
```
jbednar: It sure would be nice if .travis.yml didn't have to hard-code the name of each project, so that projects can be added and deleted without editing this shared file.

opensky/anaconda-project.yml
```diff
@@ -0,0 +1,50 @@
+# To reproduce: install 'anaconda-project', then 'anaconda-project run'
+name: opensky
+description: Datashading spatially-optimized OpenSky flight trajectories
```
jbednar: Does the spatial optimization even have any meaning for this case, as a given point won't necessarily have the destination in the same partition?

PR #27, opened 2019-05-29
Setting up website

### LINKED ISSUE #25: Set up website
One of the main things that we mean to accomplish with pyviz-topics/examples is a set of stable urls that we can link to from all of our other projects. There should be some minimal index page, but more importantly there need to be static versions of all our topics suitable for iframing and linking. Since many of these topics require large datasets and/or take a long time to run, we will want to build these things locally. I am imagining that this will look a lot like the build process for earthml, where if the project falls under either of those categories (large data or slow to run), the maintainer is responsible for uploading an evaluated version to a specific branch. Then when we use nbsite to build the website on travis, it can grab the evaluated versions from that "blessed" branch first. 

So there should be a command that the maintainer can run locally that will do this evaluation.

I guess there is also another case to think about which is the more-than-one notebook per project case. I think @jbednar's suggestion that we default to the notebook with the same name as the dir is a good one, but we might want to provide a mechanism for overriding that in the anaconda-project.yml.

I will start drafting what this will look like unless anyone has some requirements they'd like to add. @pyviz-topics/pyviz-dev

jsignell:
Closes #25 

The plan is to have separate jobs that each are in charge of generating the evaluated notebook necessary to build html using nbsite. These evaluated notebooks might need to be stored in a special branch. Then there will be another travis stage that runs a revised nbsite gallery command to build the html using the notebooks on the special branch, generate thumbnails, and create a landing page. 

The plan is to link from sites such as datashader, holoviews, and panel rather than to iframe. This is because the change in theming will be jarring even if the header, footer, and sidebar are removed.

- [x] Run each project build in a separate job
- [x] Set up travis to run a build job for only a specific tag/commit
- [x] Make a landing page that provides some toc and general info
- ~~[ ] Include versions with and without header, footer, sidebar~~
- [x] Set up a special branch where we can upload evaluated notebooks that are too big to build on CI
- [x] Make sure that travis pulls from that branch when necessary
- [ ] Investigate ways to limit redundancy in travis.yml.


jbednar:
> This is because the change in theming will be jarring even if the header, footer, and sidebar are removed.

Could we choose a neutral (gray?) palette that wouldn't jar with the various sites that will link to the examples?

doc/about.rst
```diff
@@ -0,0 +1,15 @@
+About PyViz Topic Examples
+==========================
+
+PyViz Topic Examples is home for domain-specific narrative examples using
+multiple PyViz projects. Each project is isolated and fully described:
+runnable locally and deployable to Anaconda Enterprise.
```
jbednar:
```suggestion
automatically runnable both locally and deployed on a server, e.g. in Anaconda Enterprise.
```

doc/about.rst
```diff
@@ -0,0 +1,15 @@
...
+
+This work is supported and maintained by `Anaconda
+<https://www.anaconda.com>`_.  PyViz Topics is part of the `PyViz
+<https://pyviz.org>`_ project to make Python data visualization tools
+simpler and more powerful.
+
+The primary PyViz developers are Philipp Rudiger, James A. Bednar,
+Julia Signell and Jean-Luc Stevens, with bug reports and patches from
```
jbednar:
```suggestion
Julia Signell, and Jean-Luc Stevens, with bug reports and patches from
```

doc/about.rst
```diff
@@ -0,0 +1,15 @@
...
+PyViz Topic Examples is home for domain-specific narrative examples using
+multiple PyViz projects. Each project is isolated and fully described:
+runnable locally and deployable to Anaconda Enterprise.
+
+This work is supported and maintained by `Anaconda
+<https://www.anaconda.com>`_.  PyViz Topics is part of the `PyViz
+<https://pyviz.org>`_ project to make Python data visualization tools
+simpler and more powerful.
+
+The primary PyViz developers are Philipp Rudiger, James A. Bednar,
```
jbednar:
```suggestion
The primary PyViz maintainers are Philipp Rudiger, James A. Bednar,
```

(on a site like this I'd call the authors of the notebooks the "developers"; we're just maintaining it.  (And the authors list will of course change, and should be on each notebook, so we don't need to try to cover that here.)

doc/conf.py
```diff
@@ -0,0 +1,93 @@
...
+    ('Developer Guide', 'developer_guide'),
+    ('Downloads', 'downloads'),
+    ('About', 'about')
+)
+
+html_context.update({
+    'PROJECT': project,
+    'DESCRIPTION': description,
+    'AUTHOR': authors,
+    # will work without this - for canonical (so can ignore when building locally or test deploying)
```
jbednar:
```suggestion
    # WEBSITE_SERVER is optional for local or test deployments, but provides support for canonical URLs for full deployments
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+usually start by using nbrr (``conda install -c conda-forge nbrr``)
+which reads the notebooks and looks for dependencies:
+
+.. code:: bash
+
+   nbrr env --directory "." --name bears > anaconda-project.yml
+
+**NOTE:** We tend to add ``nomkl`` to the list of dependencies to speed
+up environment build times. But there is no rule that you must do this.
+MKL is used for better runtime performance in numpy operations, since we
```
jbednar:
```suggestion
MKL is used for better runtime performance in numpy operations, but since we
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+use Numba for most of the internal computations it’s not as important.
```
jbednar:
```suggestion
use Numba for most of the internal computations it’s not as important
for these particular projects.
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+.. code:: yaml
+
+   - pip:
+     - git+https://github.com/USERNAME/REPO.git@REF#egg=PACKAGE
+
+Where ``USERNAME`` is the GitHub username, ``REPO`` is the name of the
+git repository, ``REF`` is a git reference (e.g a git tag or simply
+``master`` to point to the very latest version) and ``PACKAGE`` is the
+name of the corresponding Python package. This syntax will use pip to
+fetch the necessary code, checkout the specified git reference and
```
jbednar:
```suggestion
fetch the necessary code, check out the specified git reference, and
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+
+   anaconda-project run test
+
+You might need to declare extra dependencies or add data downloads (see
+bay_trimesh for an example of downloading data).
+
+5. For remote or large data (optional)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Make a smaller version of the data and put it in
```
jbednar:
```suggestion
Unless your data is small enough that it can be processed on every 
continuous-integration build, you should make a much smaller version
of the data and put it in
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+   ├── catalog.yml
+   └── data
+       └── f890ce4d538240e87ede9d31a6541443
+           └── data.csv
+
+Make sure to make a test catalog and put it in ``test_data/catalog.yml``
+
+7. Add the project to travis
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Once everything is setup add your project to ``.travis.yml`` following
```
jbednar:
```suggestion
Once everything is set up add your project to ``.travis.yml`` following
```

("setup", "checkout", "login", etc. are nouns; here and above we need an adjective phrase...)

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+the pattern that the other projects use.
+
+8. Upload to AE (optional)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Once you have sucessfully created the new project, you can upload and
+deploy it in Anaconda Enterprise:
```
jbednar:
```suggestion
deploy it in Anaconda Enterprise, which is the server we use to host our public Python-backed examples:
```

doc/developer_guide.rst
```diff
@@ -0,0 +1,137 @@
...
+
+::
+
+   cd bears
+   anaconda-project archive bears.zip
+
+Then in the AE interface select “Create”, “Upload Project” and navigate
```
jbednar:
```suggestion
Then in the AE interface select “Create”, then “Upload Project” and navigate
```

doc/getting_started.rst
```diff
@@ -0,0 +1,78 @@
+***************
+Getting Started
+***************
+
+This website consists of isolated fully described projects, runnable locally
+and deployable to Anaconda Enterprise.
```
jbednar:
```suggestion
and  also deployed as public examples using Anaconda Enterprise.
```

environment.yml
```diff
@@ -0,0 +1,28 @@
...
+  - bokeh ==1.1.0
+  - cartopy ==0.17.0
+  - colorcet ==2.0.1
+  - dask ==1.2.2
+  - datashader ==0.7.0
+  - fastparquet ==0.3.0
+  - geoviews ==1.6.2
+  - holoviews ==1.12.3
+  - lxml
+  - matplotlib
```
jbednar: Shouldn't mpl be pinned?  It does get some big updates...

doc/developer_guide.rst
jsignell: Those always get me :\

environment.yml
jsignell: Yes. Everything will be pinned. Just trying to get what's needed.

jsignell:
I am going to merge this once the last commit passes. The website is at https://examples.pyviz.org

PR #39, opened 2019-06-18
Set bgcolor to white for the last three taxi renderings

jlstevens:
The background color was inappropriate given the chosen color map.


jbednar:
Actually, now I'm confused.  I checked the original notebook at https://anaconda.org/jbednar/nyc_taxi/notebook , and all three were black.  So I can't tell why the colormaps are now incompatible with black?

jbednar:
The "hour of day" ones are fine, and should be kept as they are.  It's only the blue/red one that's broken, and it's not that it needs to be on white, it's that there is something wrong in the colormapping.  Original version:
image

Current version (broken):
image

jlstevens:
Turns out the colormap chosen was supposed to be against a white background for one of those plots only...

PR #42, opened 2019-06-25
Adding datashader dashboard

jsignell:
Uses data from other project dirs, so will not work properly in isolation and will need some tweaks on AE.


jsignell:
@jbednar not sure if you have an opinion about this approach of using the data that is downloaded in the other projects. It is the fastest if you are planning on running lots of these, but not the most isolated.

test_data/datashader_dashboard/catalog.yml
```diff
@@ -0,0 +1,31 @@
...
+        persist: true
+      plots:
+        dropoff:
+          label: NYC Taxi Dropoffs
+          x: dropoff_x
+          y: dropoff_y
+        pickup:
+          label: NYC Taxi Pickups
+          x: pickup_x
+          y: pickup_y
```
jbednar: Please set your editor to enforce newlines at the end of files...

jsignell: good call.

PR #43, opened 2019-06-25
Numba compatibility fixes for attractor notebooks

jlstevens:
Latest numba encounters [an issue](https://github.com/numba/numba/issues/4223) in the current clifford notebook due to the creation of a pandas dataframe in the function return.

One fix would be to pin back numba but the better approach is to make nopython mode explicit and create the dataframe outside the jitted section. This PR uses this approach (which is being encouraged by the numba devs) to fix the three notebooks in this project.


attractors/attractors_panel.ipynb
```diff
@@ -26,13 +26,17 @@
...
+    "@jit(nopython=True)\n",
+    "def trajectory_coords(fn, x0, y0, a, b, c, d, e, f, n):\n",
     "    x, y = np.zeros(n), np.zeros(n)\n",
     "    x[0], y[0] = x0, y0\n",
     "    for i in np.arange(n-1):\n",
     "        x[i+1], y[i+1] = fn(x[i], y[i], a, b, c, d, e, f)\n",
-    "    return pd.DataFrame(dict(x=x,y=y))"
+    "    return x, y\n",
+    "\n",
+    "def trajectory(fn, x0, y0, a, b=None, c=None, d=None, e=None, f=None, n=1000000):\n",
```
jbednar: Using a,b,c,d,e,f as arguments for all functions was to match Numba's requirement for a fixed argument list. Now that we are using two functions already, it might be better to accept *args here in this function's argument list, and then raise an error if there are too many.  But it's probably not worth worrying about; I'm just recording that sentiment here for posterity.

PR #46, opened 2019-06-27
Renaming dashboard.ipynb -> datashader_dashboard and adding logo

jsignell:


jbednar:
Can you please merge the changes from https://github.com/pyviz/datashader/pull/768, maybe waiting for hvplot to be released so that it works?

jsignell:
Sure, but hvplot isn't released yet, so I don't think the time is right.

jbednar:
Right. In the meantime, doesn't intake need to be pinned here to avoid the problems that the datashader PR addresses?

jsignell:
I got rid of the ranges in the catalog. It doesn't make a big difference in my opinion.

jbednar:
Ah, that works too. :-)

PR #47, opened 2019-07-02
Fixing up glaciers logo

jsignell:


jbednar:
Looks good. @philippjfr may prefer using the Panel logo instead?

philippjfr:
I would, makes more sense to me.

philippjfr:
Alternatively we drop it entirely. There's already the oggm logo there.

jbednar:
I think having the Panel logo on there is good, as there is a Dash version (which we link to) that can have the Dash logo.

PR #49, opened 2019-07-18
Don't link to local assets. Use urls instead for images

jsignell:
It is tricky to make sure that assets end up in the right places. Better to just use urls to resources on s3.


jbednar:
We keep switching back and forth about this.  The problem with url-based access is that the example can't run behind a firewall. :-(

jsignell:
Yes, but hauling around assets for each of the projects is a total pain. It would need to be implemented in nbsite I guess so that the links to the assets can be swapped out with ones that are up a level. Or we can embed the images as base64, but that means they can't be as big.

Otherwise we'd need to add a top level assets dir, but that undermines the isolation between the projects.

jsignell:
We already depend on tiles all over the place so I think internet access is a prerequisite

jbednar:
The difference is that when there's no internet access, examples with map tiles still work otherwise; there just aren't any map tiles shown.  But when a Panel app (or Jupyter notebook?) refers to an image by URL, it aborts entirely.  Maybe we should make Panel simply skip the image or substitute a broken image when it's not accessible, so that the app remains otherwise usable when offline?

jbednar:
This happened to me just a couple of days ago on an airplane, where I kept having to comment out the image when running anything, and kept forgetting to uncomment it when committing a new version to the repo.  But even more importantly, I want to be sure that people behind a firewall can still see how things are meant to work.

jsignell:
Using approach in #53 instead

PR #51, opened 2019-07-29
Adding labels to anaconda_project.yml and site. 

### LINKED ISSUE #50: Show which projects use only packages from defaults
There has been a discussion around how to expose which projects work with only the defaults channel. 

This work has two parts:
1) Going through the projects in this repo and removing the pyviz channel dependency to see whether they can build with just defaults (estimate 2 days).

2) Displaying that in the static version of the site. 
  
      a. I think the right way of doing that would be to implement a labelling system which would then show up on the gallery page or generate new gallery pages for each label. This would requite changes in nbsite and making nbsite aware of anaconda-project.yml in a way that it isn't yet. But it would also allow us to label projects with 'datashader' or 'geoviews' which serves our long-term goals (estimate 4 days).

     b. Instead of implementing the label system, we could cheat and make a .rst with static links to the right gallery assets (estimate 1 day).

jsignell:
Closes #50

Screen Shot 2019-07-31 at 2 23 47 PM


jsignell:
Note that I am using nbsite straight from github since anaconda.org is being a little flakey right now.

PR #52, opened 2019-07-30
Made channel URL explicit

philippjfr:


jbednar:
What is the motivation for this change?  It has a very serious downside, which is that it prevents these examples from being used inside an airgap or firewall where the `pyviz` channel is mirrored locally but internet access is blocked.  Ideally most of these examples will soon be rewritten to depend only on defaults, which will avoid the issue, but omitting the URL was a deliberate policy that should only be changed for a reason that outweighs the problems it causes.

philippjfr:
Read the discussion on slack.

PR #67, opened 2019-10-09
WIP: Widget-based Datashader dashboard

jbednar:
This PR adds a version of the Datashader dashboard based on explicit widget objects rather than defining a class with Parameters from which widgets are synthesized.  The functionality should be identical with HoloViews >= 1.12.6 and Panel >= 0.6.4, but it shows the user how to specifically control the type and parameters of every widget while not requiring users to learn about Parameters or about writing their own Parameterized (or any other) custom classes. The two dashboards are meant to be so closely aligned that `diff` can be used to see the specific differences between the two approaches.  

What makes this still be WIP is that when you uncomment `#pn.Row(widgets,view)` in `widgetdashboard.ipynb`, the output of that cell doesn't show any plot.  It does show a plot if you change the return value of `view()` to just `dataplot * labels` (no map tiles), but including the map tiles or their options messes it up.  @philippjfr , any ideas?

As it's just a demo, `widgetdashboard.ipynb` also defaults to `nyc_taxi_50k`, not `nyc_taxi`, but once it works it can be changed to share the same default as `dashboard.ipynb`.

### DIFF
--- a/datashader_dashboard/dashboard.ipynb
+++ b/datashader_dashboard/dashboard.ipynb
@@ -79,7 +79,7 @@
     "cmaps  = odict([(n,colorcet.palette[n]) for n in ['fire', 'bgy', 'bgyw', 'bmy', 'gray', 'kbc']])\n",
     "\n",
     "maps   = ['EsriImagery', 'EsriUSATopo', 'EsriTerrain', 'CartoMidnight', 'StamenWatercolor', 'StamenTonerBackground']\n",
-    "bases  = odict([(name, getattr(hvts, name)().relabel(name)) for name in maps])\n",
+    "bases  = odict([(name, hvts.tile_sources[name]().relabel(name)) for name in maps])\n",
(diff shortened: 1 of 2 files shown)


jsignell:
Adding a height to the base opts seemed to fix the issue.

philippjfr:
Can it also be fixed with a min_height? That way responsive behaviour isn't entirely disabled.

jsignell:
Yeah that is a better solution.

jbednar:
But isn't that just a workaround? Why should including map tiles affect whether there is a plot?

jsignell:
It doesn't affect whether there is a plot. It affects whether the plot has height=0 presumably because having responsive set to True trumps height, width set on the other plot. 

Maybe the question is should responsive=True have a default min_height? I imagine that has been discussed at length in a different repo.

maximlt:
@jbednar do you still want this new notebook to be part of the examples? Asking as the PR is pretty old, you might have changed your mind ;)

jbednar:
Yes, this should be updated and merged, particularly since it's the approach I'd recommend to most users starting out. I sure hope that there are answers to the above issue now!

PR #74, opened 2019-11-20
Reorganizing docs to make them more findable

jsignell:
Also adding references to build options like skip and orphans and explaining how index works.


doc/maintenance.rst
```diff
@@ -0,0 +1,76 @@
...
+The primary PyViz maintainers are Philipp Rudiger, James A. Bednar,
+Julia Signell, and Jean-Luc Stevens, with bug reports and patches from
+numerous members of the Github community.
+
+Website Maintenance
+===================
+The website building occurs in travis jobs that are triggered by commit messages.
+
+**NOTE:** When you merge PRs, you should be very careful about what is included
+in the commit messages.
```
jbednar: That sounds scary; can you explain what to watch out for?

jsignell: Oh just that you aren't triggering unnecessary builds. Maybe I should take out the "very".

jbednar: Just say something to explain what you mean...

Issue #76, opened 2019-11-26
Portfolio optimizer upload csv raises exception


jordansamuels:
I'm running the portfolio optimizer off of the files downloaded from https://examples.pyviz.org/assets/portfolio_optimizer.zip .  Everything appears to work fine, except when I upload a csv of stock prices, I get the stack trace below.  This does not happen when I use https://portfolio-optimizer.pyviz.demo.anaconda.com/portfolio, which works fine with the same csv file.  I also get the same error when I run from source (abc016730c8ccf04a0a2fef35f4232e3f2ec9468).  

I'm using 
```
% conda --version; anaconda-project --version
conda 4.7.12
0.8.3
```
on OSX.  

Exception trace:
```
2019-11-25 09:42:26,235 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7fd1e0572a58>>, <Future finished exception=RuntimeError('FileInput.filename is a readonly property',)>)
Traceback (most recent call last):
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/tornado/gen.py", line 748, in run
    yielded = self.gen.send(value)
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/bokeh/server/session.py", line 70, in _needs_document_lock_wrapper
    result = yield yield_for_all_futures(func(self, *args, **kwargs))
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/bokeh/server/session.py", line 191, in with_document_locked
    return func(*args, **kwargs)
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/bokeh/document/document.py", line 1127, in wrapper
    return doc._with_self_as_curdoc(invoke)
  File "/private/tmp/portfolio_optimizer/envs/default/lib/python3.6/site-packages/bokeh/document/document.py", line 1113, in _with_self_as_curdoc
    return f()
[...]

PR #77, opened 2019-11-26
Fix portfolio_optimizer exception (#76) by upgrading panel, bokeh

jordansamuels:
Based on suggestion by @philippjfr in discussion in pyviz Gitter.


philippjfr:
Thank you, has been fixed separately in the meantime.

PR #83, opened 2019-12-17
Added voila_gpx_viewer

jbednar:
Added example by @jtpio from https://github.com/jtpio/voila-gpx-viewer/commit/70d1f64a .

Deployed at https://voila-gpx-viewer.pyviz.demo.anaconda.com/


jbednar:
It looks like evaluation (and therefore publishing) of this project is failing due to conflicting requirements for nbconvert between voila and nbsite:

voila=0.1.9 -> nbconvert[version='>=5.5.0,<6']
nbsite=0.6.5 -> nbconvert[version='<5.4']

It may be possible to roll back voila for this particular example, but in general people are going to want the most recent versions of a fast-changing project like voila, so I think we need to be able to support later nbconvert versions.

jtpio:
Thanks @jbednar for adding this example!

There was no particular reason to use `0.1.9`, so I've just updated `voila` to the latest version (`0.1.20`). Maybe that could help make the deployment smoother?

jbednar:
The problem was in the other direction; the `nbsite` command we were using was requiring an nbconvert version too old, not too new, so updating voila further would have only made it worse if anything.  We fixed it on the `nbsite` side, and it's now fine.  This example is now deployed at https://examples.pyviz.org/voila_gpx_viewer/voila_gpx_viewer.html , alongside with lots of others using other libraries. Thanks for creating it!

PR #86, opened 2020-01-15
Annotator overlays in ml_annotators

jbednar:
Using the changes in https://github.com/holoviz/geoviews/pull/429 and https://github.com/holoviz/holoviews/pull/4184, it should now be possible to create an overlay and then pass that in to an annotator, as long as the overlay has only one annotable element.  That capability makes it much easier to explain and illustrate how annotators work for overlaid data, because it lets us avoid explaining the confusing `annotate.compose` method until it is actually needed (in the case of multiple annotatable objects in the same overlay).  

Assuming those PRs are completed and merged, this PR now introduces drawing tools on their own, using an overlay, and only then shows how that overlay can be used in an annotator. This approach helps introduce one concept at a time and let the reader keep everything straight.

To do:

- [ ] The overlay support in https://github.com/holoviz/holoviews/pull/4184 is not yet ready; needs additional work before this notebook will run without errors and show the overlay. 
- [ ] The two linked PRs need to be merged and dev releases generated before this PR can be merged
- [ ] The changes to the .yml file should be validated against the actual value of the hv and gv dev releases needed; they were bumped by 1 but may not match where the actual changes end up
- [ ] After the above changes are done, need to test this notebook to make sure that overlays show up everywhere and that the descriptions are accurate
- [ ] How a user works with the PointDraw stream is different from how they work with the PointsAnnotator (e.g. annotator.annotated.dframe() vs pd.Dataframe(stream.data), no Shapely support that I can see, etc. Are all of those differences intentional and unavoidable?  If not it seems like it would be nice for people to be able to switch between the drawing-tool stream and the annotator as needed depending on whether they need an attached table or not, without seeing the APIs have to change entirely in the two cases. This is low priority, but worth considering.
- [ ] Need to point to annotate.compose somewhere and mention that things get more complicated when overlaying multiple annotatable elements.

If it all works after the above, it should be merged and the running instance updated.

### DIFF
--- a/ml_annotators/anaconda-project.yml
+++ b/ml_annotators/anaconda-project.yml
@@ -18,8 +18,8 @@ packages: &pkgs
   - ipykernel=5.1.0
   - pandas=0.24
   - bokeh=1.4.0
-  - holoviews=1.13.0a21
-  - geoviews=1.7.0a6
+  - holoviews=1.13.0a22
+  - geoviews=1.7.0a7
   - panel=0.7.1a2
 
 dependencies: *pkgs
--- a/ml_annotators/ml_annotators.ipynb
+++ b/ml_annotators/ml_annotators.ipynb
@@ -6,7 +6,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import holoviews as hv\n",
+    "import holoviews as hv, holoviews.streams as hvs\n",
     "import geoviews as gv\n",
     "\n",
     "hv.extension('bokeh')"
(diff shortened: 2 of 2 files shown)


maximlt:
> How a user works with the PointDraw stream is different from how they work with the PointsAnnotator (e.g. annotator.annotated.dframe() vs pd.Dataframe(stream.data), no Shapely support that I can see, etc.). Are all of those differences intentional and unavoidable? If not it seems like it would be nice for people to be able to switch between the drawing-tool stream and the annotator as needed depending on whether they need an attached table or not, without seeing the APIs have to change entirely in the two cases. This is low priority, but worth considering.

I'm not sure this is still valid, but if it is it would be better to open an issue where relevant as it can easily get lost here :)

Otherwise this looks like good improvements that could be merged in, after resolving the last TODO items and rebuilding the project.

PR #87, opened 2020-01-31
Adding Walker Lake project

jlstevens:
This project is the [Walker Lake example](https://earthml.holoviz.org/topics/Walker_Lake.html) from earthml.holoviz.org. This project both updates the necessary dependencies, makes a number of fixes needed for reproducibility and makes sure only the dependencies that are needed are listed (on contrast to the full EarthML environment that no longer works properly).

I don't think this project should be more than an HTML build of the notebook: a read-only notebook deployment would take quite a long time to run and I am not sure how much more useful it is than a simple snapshot.

### FIXES:

- [x] The AWS S3 region wasn't specified which meant intake could not reliably fetch the necessary data.
- [x] Pinned `distributed` to `2.9.3` as the new `2.10.0` version causes the initial (datashaded) HoloMap to hang.
- [x] Silenced the dask performance warnings issued while rechunking.

### TODO:

- [ ] Investigate the `distributed 2.10.0` issue further.
- [ ] Remove dependence on EPSG server. There were problems communicating with the server earlier this week and it would be good to remove a potential point of failure.
- [ ] Capture the full lock of the project environment.
- [ ] Hook up with the `landsat5_small` and `landsat8_small`  datasets for testing.
- [ ] Add a warning to the top of the notebook stating how long it should take to run the whole notebook?

PR #89, opened 2020-02-05
Walker lake fixes

jlstevens:
This PR addresses the threading issue described in #87. It was rather tricky to track down but here is the summary:

1. Occasionally displaying the rasterized holomap would hang and when interrupted, the traceback indicated that some sort of deadlock was occurring. Sometimes the output would appear, sometimes the output would hang.

2. This appeared to be due to an interaction between the numba parallel code used in regridding and dask.

3. Installing `tbb` is recommended for numba as it is a thread safe scheduler while the default one is not. I added it to the `anaconda-project.yml` but even with `tbb` installed, the issue persisted.

4. It turns out the use of dask in this notebook was not appropriate anyway: it is better to have multiple single-threaded dask workers than one multi-threaded dask worker, at least when parallel numba code is involved. The former configuration resolves the hanging issue.

5. It turns out that even though `tbb` was installed, the issue was that `tbb` support in the numba package on conda forge was *not* enabled (at least on OSX, where I am testing things locally). There is an issue about this: https://github.com/conda-forge/numba-feedstock/issues/43 and hopefully the conda-forge package will soon be fixed in this regard.

Glad that is all finally figured out! From the sound of it, we should always install `tbb` (at least once the conda-forge numba package is updated) whenever we are using datashader and dask together.


philippjfr:
Nice, glad you tracked this down!

walker_lake/Walker_Lake.ipynb
```diff
@@ -48,10 +48,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# arbitrarily choose a small memory limit (4GB) to stress the \n",
+    "# arbitrarily choose a small memory limit (4GB) to stress the\n",
     "# out of core processing infrastructure\n",
     "from dask.distributed import Client\n",
-    "client = Client(memory_limit=10e10, processes=False) # Note: was 6e9 \n",
+    "client = Client(memory_limit=10e10, n_workers=8, threads_per_worker=1) # Note: was 6e9 \n",
```
jbednar: Seems like these messages need updating; neither 10e10 nor 6e9 are 4GB, so please make the value match the message or the message match the value ("100GB"?).

jbednar: Also please add a comment above "client" saying that because Datashader uses parallel Numba for raster rendering, we tell Dask to use only a single thread on a given CPU to avoid contention.

jlstevens: Addressed in ab5cd76

jlstevens: Your second comment only just appeared for me. I'll add that description shortly.

jlstevens: Comment added in a39c919. I think this is probably worth some markdown explanation as well tbh so I'll add that now.

walker_lake/Walker_Lake.ipynb
```diff
@@ -48,10 +59,12 @@
...
    "source": [
-    "# arbitrarily choose a small memory limit (4GB) to stress the \n",
-    "# out of core processing infrastructure\n",
+    "# arbitrarily choose a memory limit (8GB) to demonstrate the out of core\n",
+    "# processing infrastructure\n",
     "from dask.distributed import Client\n",
-    "client = Client(memory_limit=10e10, processes=False) # Note: was 6e9 \n",
+    "client = Client(memory_limit=8*1e9, n_workers=8, threads_per_worker=1)\n",
+    "# As Datashader uses parallel Numba for raster rendering, we need to use\n",
+    "# single threaded Dask workers to on each CPU to avoid contention.\n",
```
jbednar:
```suggestion
    "# single threaded Dask workers on each CPU to avoid contention.\n",
```

jlstevens:
Getting the Walker Lake notebook to work with sensible test data was trickier than expected:

1. First the original 'small' data was inappropriate as it consisted of slices of the original data with different bounds. This meant a lot of the values used in the notebook were inappropriate.
2. The solution was to downsample the original dataset by a factor of 100 in both dimensions (I used QGis for this)
3. Using this downsampled test data revealed some other issues, the use of xarray `.interp()` to downsample stop working (for reasons I do not yet understand). The solution was to use datashader to resample appropriately.
4. Some of the things that *seemed* to be working were in fact very suspect. For instance, the `collapse` operation with `np.sum` was working when the two xarrays had different shapes. @philippjfr suggests we should put checks in holoviews to disable this xarray behavior (which apparently merges the coordinates).
5. Now I resample right at the start to make sure the 1988 and 2017 data is the same shape.

It is now working locally both for the real and test data sets. Fingers crossed it will now pass on Travis!

jlstevens:
Walker Lake finally built on Travis! Unfortunately, the output isn't what we want though: at the end of the notebook, the node ran out of memory resulting in a traceback. This also meant the thumbnail wasn't generated either.

I will now see if there is some way to work around this, ideally without having to build things locally. Otherwise, I think this PR is working as well as it can given the hardware limitations of the build infrastructure.

jbednar:
What a long journey this has been! Seems to work now, so merging.

PR #95, opened 2020-02-17
Updated conf.py to use new deployment_url semantics

jlstevens:
This PR aims to test recent changes to nbsite (in the latest dev release) to see if the deployment links are now correctly generated for the IEX project (I'll be building that project in particular as the links are currently incorrect).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -261,7 +261,7 @@ jobs:
         - git checkout -b deploy-${TRAVIS_BRANCH}
         - git fetch https://github.com/$TRAVIS_REPO_SLUG.git evaluated:refs/remotes/evaluated # all evaluated notebooks, json, and rst should be checked in to this branch
         - git checkout evaluated -- ./doc
-      script: doit build_website
+      script: ls -al .; git status; doit build_website
       deploy:
         - provider: pages
           skip_cleanup: true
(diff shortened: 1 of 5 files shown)


jlstevens:
@jbednar The links now look right in the last build. I can't be 100% sure the behavior is now correct for all projects but it does now work for IEX.

jlstevens:
This PR became a combined mess and nightmare! To summarize what went wrong and how I fixed it:

1. I added `user_fields` to the `anaconda-project.yml`.
2. I didn't pop this field in `dodo.py` which caused `anaconda-project.yml` to be dirty on Travis.
3. This meant Travis couldn't checkout the `evaluated` branch.
4. This in turn meant that at the end of the build, Travis push forced this PR to `evaluated`, entirely removing all reference to the evaluated content from all the projects!
5. I had to fish around using the curl commands described in [this stackoverflow post](https://stackoverflow.com/questions/12568628/how-to-recover-from-a-git-push-force) to guess a good SHA from the GitHub reflog and create a new branch from the orphan. Somehow I got lucky and found a reference to a commit before the force!
6. Now I can close this PR and pretend none of this ever happened. :-)

Travis should really have aborted when the checkout failed instead of continuing onto the push force. I'll have to add some bash logic to do this manually.

PR #96, opened 2020-02-21
Fixes to evaluated branch and deployment links

jlstevens:
Takes the good bits of #95 and makes sure the dev website is fixed now the evaluated branch is recovered.

### DIFF
--- a/iex_trading/anaconda-project.yml
+++ b/iex_trading/anaconda-project.yml
@@ -14,6 +14,11 @@ labels:
 skip:
   - IEX_to_CSV.ipynb
 
+user_fields:
+  - labels
+  - skip
+  - maintainers
+  
 channels: [pyviz/label/dev]
 
 packages: &pkgs
--- a/iex_trading/anaconda-project.yml
+++ b/iex_trading/anaconda-project.yml
@@ -14,10 +14,7 @@ labels:
 skip:
   - IEX_to_CSV.ipynb
 
-user_fields:
-  - labels
-  - skip
-  - maintainers
+user_fields: [labels, skip, maintainers]
   
 channels: [pyviz/label/dev]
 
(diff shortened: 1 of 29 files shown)


jlstevens:
In the end, this PR does these things:

* Updated `anaconda-project`
* Uses `user_fields` supported by this version of `anaconda-project`
* Only installs holoviews for website building if it is not already available.
* Use the new nbsite build to create the correct deployment links (original purpose of this PR)

@jbednar Ready to merge if you agree that the deployment links on  https://pyviz-dev.github.io/examples/ are now appropriate (all the ones I checked seemed fine to me)

iex_trading/anaconda-project.yml
```diff
@@ -14,6 +14,11 @@ labels:
 skip:
   - IEX_to_CSV.ipynb
 
+user_fields:
```
jbednar: I think this should be made more compact because it will be in every file, right?

```
user_fields: [labels, skip, maintainers]
```

And it also needs copying to the other .yml files, right?

jlstevens:
After lots of confusion, I ended up leaving the `anaconda-project` version alone due to the issue described in https://github.com/Anaconda-Platform/anaconda-project/issues/255. In short, I've added `user_fields` to all the projects which should be supported in 0.8.4 but it ends up modifying the `anaconda-project.yml` during prepare (for some unknown reason) which then causes the checkout of the evaluated branch to abort.

jlstevens:
The root cause has now been identified. Anaconda project 0.8.4 has a bug fixed in https://github.com/Anaconda-Platform/anaconda-project/pull/256 whenever `user_fields` was used. This means we will want to update `anaconda-project` once the next release is available (presumably 0.8.5?).

jlstevens:
@jbednar The tests have now gone green again. I think this PR is ready to merge noting that the following issues still need to be tackled in subsequent PRs:

- [ ] The `opensky` and `gerrymandering` projects had test failures when their `anaconda-project.yml` files had `user_fields` added. This will need to be investigated.
- [ ] The `user_fields` field won't be respected till `anaconda_project 0.8.5` is released and used in `dodo.py`.
- [ ] There is still some issue around param pinning to get rid of the API deprecation warnings.

I think this PR is worth merging if only to get the deployment link fixes into master ASAP.

PR #98, opened 2020-02-24
Implementing safe push to evaluated branch on Travis

jlstevens:
This PR aims to avoid the problem when the evaluated branch fails to checkout, resulting in a bad push force that overwrites the branch.


jlstevens:
This PR is a one line change in the end but it has now been tested and seems to be working as intended, protecting us from push forcing when the checkout fails. @jbednar Ready to merge (though you may want to wait for the tests to go green first if you like).

PR #103, opened 2020-03-31
Converted Seattle Lidar example to a project

jlstevens:
Converts the [Seattle Lidar example on EarthML](https://earthml.holoviz.org/topics/Seattle_Lidar.html) to a project. Testing and running correctly locally. This project might be able to work on Travis as long as the intake download from S3 can happen fast enough. When the necessary data is in the cache, the notebook runs fairly quickly.

PR #104, opened 2020-03-31
Converted Carbon Flux example to a project

jlstevens:
Converts the [Carbon Flux example on EarthML](https://earthml.holoviz.org/topics/Carbon_Flux.html) into a project. Tested and working correctly locally. Just as with the #103, the reason it might timeout on Travis is that downloading all the fluxnet data can be quite slow if it isn't cached.


jlstevens:
@jbednar I'm going to merge this PR so I can build `landsat_clustering`. Happy to open a new PR to address any suggestions you might have.

PR #107, opened 2020-04-10
updating the html theme to use sphinx_holoviz_theme

kebowen730:
@philippjfr @jlstevens


kebowen730:
Unnecessary theme update

PR #116, opened 2020-06-17
Updating .travis.yml to use standardized build tags

kebowen730:
This PR updates the accepted tags for triggering website builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ stages:
   - name: build_project
     if: commit_message =~ /^.*(build:).*$/
   - name: website_dev
-    if: (commit_message =~ /^.*(build:).*$/) OR (branch = master)
+    if: tag =~ ^v(\d+|\.)+[a-z]\d+$ OR tag = website_dev
   - name: website_release
-    if: (commit_message =~ /^.*(build:).*$/) AND (commit_message =~ /^.*(website_release).*$/)
+    if: tag =~ ^v(\d+|\.)+[^a-z]\d+$ OR tag = website
 
 jobs:
   include:


.travis.yml
```diff
@@ -22,9 +22,9 @@ stages:
   - name: build_project
     if: commit_message =~ /^.*(build:).*$/
   - name: website_dev
-    if: (commit_message =~ /^.*(build:).*$/) OR (branch = master)
+    if: tag =~ ^v(\d+|\.)+[a-z]\d+$ OR tag = website_dev
   - name: website_release
-    if: (commit_message =~ /^.*(build:).*$/) AND (commit_message =~ /^.*(website_release).*$/)
+    if: tag =~ ^v(\d+|\.)+[^a-z]\d+$ OR tag = website
```
jbednar, line 27: 1. It would be easier for us to remember if the tag name (`website`) and the task name (`website_release`) were the same; I vote for consistently calling both the tag and the task `website_dev` and `website_release` so that they are completely clear about which website it is.
2. It's adding support for tagging a versioned release, but this particular repo does not have versioned releases (it's a grab bag of separately versioned examples), so adding that is optional. If it helps this site be more comparable to our other sites, sure!
3. This PR seems to be removing support for building by commit message (`git commit -m 'Updated website [build:website_dev]'), which is a workflow that's more useful than git tagging except for actual releases. So please restore that support unless it has somehow moved elsewhere.

kebowen730: 1. I can do that, I was initially standardizing off of what seemed to be the prevailing tag convention in a few other repos, I'll go back through the other repos and update them as well.

2. I can take that out, I initially added it going off of other .travis.yml file, unless you think it's worth keeping in.

3. I can revive that capability, but does that just apply for this repo? I think there are a couple others that allow for commit messages as well, but I removed that capability there too because I thought the intent was to standardize on using tags for website releases.

jbednar: 1. Thanks!
2. Probably worth keeping solely for consistency
3. All repos should consistently support building on commit messages; we want to standardize the specific tags used, but more often (except for actual releases) we will use commit messages to control builds rather than using tags.

philippjfr:
No longer using Travis.

PR #131, opened 2021-01-26
Drop spatial indexing (for now)

philippjfr:
The census and osm parquet files currently use the old datashader.spatial spatial indexing which has been removed. While we decide whether to update them to use spatialpandas this at least fixes these notebooks.


census/census.ipynb
```diff
@@ -32,16 +31,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%time\n",
-    "df = spatial.read_parquet('./data/census2010.parq')\n",
-    "df = df.persist()"
+    "df = dd.read_parquet('./data/census2010.parq/', engine='fastparquet').persist()\n",
```
jbednar, line 34:
```suggestion
    %%time
    "df = dd.read_parquet('./data/census2010.parq/', engine='fastparquet').persist()\n",
```

As a general rule, I put `%%time` on any cell that takes more than a few seconds to run, or that a user might _wonder_ whether it takes more than a few seconds to run. Here, reading the file is usually only a few seconds, but if it were CSVs it would take minutes, so I think it's important to convey to someone reading the notebook when rendered on the website that this step should take e.g. 8 seconds.  They might think it would take 1 second or 1 minute, so showing them explicitly what it takes is helpful.

census/census.ipynb
```diff
@@ -32,16 +31,18 @@
...
+    "\n",
+    "df['race'] = df.race.astype('category').cat.as_known()"
```
jbednar, line 36: Why does it specify the engine as fastparquet if it then sets the category?  I thought the only reason we require fastparquet is that it preserves the category, so if we're converting categories anyway should it use pyarrow instead?

census/census.ipynb
```diff
@@ -32,16 +31,18 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Here we call ``.persist()`` to use Dask's fast in-core operations, but if you have less than 16GB of RAM, you can omit that line to use Dask's support for out-of-core (larger than memory) operation. Working out of core will be much slower, but should work even on small machines."
+    "Here we call ``.persist()`` to use Dask's fast in-core operations, but if you have less than 16GB of RAM, you can omit that line to use Dask's support for out-of-core (larger than memory) operation. Working out of core will be much slower, but should work even on small machines.\n",
```
jbednar, line 43:
```suggestion
    "Here we call ``.persist()`` to use Dask's fast in-core operations, but if you have less than 16GB of RAM, you can omit that call to use Dask's support for out-of-core (larger than memory) operation. Working out of core will be much slower, but should work even on small machines.\n",
```

(.persist() is no longer on a separate line)

census/census.ipynb
```diff
@@ -611,8 +616,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "population = dynspread(datashade(points, cmap=fire, element_type=gv.Image))\n",
-    "gts.EsriImagery() * population * gts.StamenLabels().options(level=\"annotation\")"
+    "population = dynspread(rasterize(points, cmap=fire, aggregator='count')).opts(cnorm='eq_hist')\n",
```
jbednar, line 619:
```suggestion
    "population = dynspread(rasterize(points, cmap=fire, aggregator='count')).opts(cnorm='eq_hist', tools=['hover'])\n",
```
(Now that it uses rasterize, should it also enable hover so that we can see the actual population counts per pixel?)

jbednar (commented):
Looks great, thanks!!

census/census.ipynb
philippjfr: Specifying the engine was my attempt at forcing it to load the categories but it doesn't seem like they were saved to this file or that no longer works. In either case, should remove the engine now that I know that.

census/census.ipynb
philippjfr: Yep, my mistake.

Azaya89:
I believe we can close this PR now based on #265 and the current #430

PR #134, opened 2021-01-29
Fixing symlink creation on GH Actions

jlstevens:
The symlinks are necessary to allow the shortened URLs on examples.pyviz. Creation was working on Travis but no longer seems to be working on GH Actions. The public website can be fixed by manually pushing the symlinks to `gh-pages` but that is easily forgotten and not a long term solution.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -49,6 +49,7 @@ jobs:
           conda activate test-environment
           doit build_website
           doit index_symlinks
+          sudo chown -R $(id -u):$(id -g) ./builtdocs   
       - name: Deploy dev
         uses: crazy-max/ghaction-github-pages@v2
         if: "!contains(github.event.head_commit.message, 'website_release')"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,7 +48,7 @@ jobs:
           eval "$(conda shell.bash hook)"
           conda activate test-environment
           doit build_website
-          doit index_symlinks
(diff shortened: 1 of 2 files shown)


jlstevens:
I now think the symlink creation is fine and it is the `crazy-max/ghaction-github-pages` step that is complaining.

Probably related to this issue: https://github.com/crazy-max/ghaction-github-pages/issues/130

jlstevens:
@jbednar I am exploring a non-symlink based approach to redirecting with pages like this:

```html
<!DOCTYPE html>
<html>
   <head>
      <title>Square Limit Redirect</title>
      <meta http-equiv = "refresh" content = "0; url = https://examples.pyviz.org/square_limit/square_limit.html" />
   </head>
   <body>
      <p>Redirecting...</p>
   </body>
</html>
```

Right now you can test this approach [here](https://examples.pyviz.org/square_limit/square_limit.html).

I'll obviously have to update `dodo.py` to use this if I can't figure out how to get the symlinks working again soon.

jlstevens:
This change has been successfully deployed to https://examples.pyviz.org/.

One way in which symlinks were better is the URL didn't change when using the short URL. For instance, if you enter `https://examples.pyviz.org/ship_traffic/` you get directed to `https://examples.pyviz.org/ship_traffic/ship_traffic.html`. This doesn't seem too bad and it is a lot better than broken links so I'll go ahead and merge. We can try to figure out how to get the symlinks updated on gh-pages in a separate PR.

PR #149, opened 2021-05-23
Updated landsat, nyc_buildings, and pyviz refs

jbednar:
The landsat example was quite out of date, with warnings from using old HoloViews syntax and tile definitions from outdated Bokeh versions, UI bugs since improved in Bokeh, and using regrid rather than rasterize (which is needlessly confusing). Updated to the current stack.

The nyc_buildings example no longer needs dev packages in most cases, but it does still depend on spatialpandas-0.4.0a1. Updated as much as I could. @philippjfr, do you know anything about the status of spatialpandas 0.4.0? 0.4.0a1 was four months ago now, so presumably we should be making a proper 0.4.0 release of spatialpandas?

Also fixed a few stray references to an older usage of PyViz that were probably confusing people.


jlstevens:
All the changes look good, I folded your `nyc_buildings` changes into #150 and I've run the new landsat deployment and it works fine. I'll merge so I can rebase #150 and separately make a PR updating the `anaconda-project.yml` format to ensure it locks properly.

PR #163, opened 2021-07-16
Updated attractors.ipynb to use hvplot and Panel

jbednar:
As an exercise in our new [hvPlot-centric approach to using our tools](https://github.com/holoviz/hvplot/issues/533), I've updated attractors.ipynb to use  hvplot and Panel instead of hv and DynamicMap. The goal isn't necessarily to make the code shorter, though it did reduce it slightly, it's to make the concepts easier to convey given what people already know about Pandas and how widgets work.

I *think* it's succeeded at that?  Certainly the first couple of .hvplot calls are straightforward; there is data in a df, and we call hvplot on it. 

The last example is less of a clear win. We can't use .interactive here because we're generating a new DataFrame each time, so I've used Panel and pn.bind. The old one only relied on HoloViews concepts throughout, from hv.Points to DynamicMap, which means that there was one place to go for reference, but even so the ideas were quite esoteric and surprising to people.  The new one adds Panel widgets and pn.bind as concepts, which they do have to figure out, but I think they are easier to convey and it's also much, much clearer where to go from here. The user can see that if they wanted to put anything else on a widget, they can just define one and hook it in, whereas with DynamicMap it was unclear how to add a Boolean, Selector, etc.

The new code in that last cell is also somewhat dense, but I can't think of any easy way to simplify it without making it much, much more verbose.

### DIFF
--- a/attractors/anaconda-project-lock.yml
+++ b/attractors/anaconda-project-lock.yml
@@ -17,7 +17,7 @@ locking_enabled: true
 env_specs:
   test:
     locked: true
-    env_spec_hash: f4754db9d2eb2686248ed993d9d52f6117c02b5a
+    env_spec_hash: c05aeba8ddbcf58853b8df42513c3b784153545c
     platforms:
     - linux-64
     - osx-64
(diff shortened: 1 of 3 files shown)


Azaya89:
This PR #381 aims to complete this PR. Maybe this one can be closed?

maximlt:
@Azaya89, let us know if there's anything in this PR that should be added to https://github.com/holoviz-topics/examples/pull/381 (which I have merged too quickly, just assume it's still open :) ). If you say no, we are going to close it.

PR #169, opened 2021-08-07
Refactored exoplanets into modular plotting functions

jbednar:
The exoplanets example looked great, but I kept wanting to be able to see plots on their own, which was difficult when they were all bound together into the `gal_map` function. I've split it into separate plotting functions for each of the items in the overlay, which makes it possible to see how everything fits together and also makes it much easier to e.g. change the format into a layout instead of an overlay if desired. This also let me put all the Panel-dashboard stuff at the very end, cleanly separating the underlying data-manipulation code from the widgets and layouts.

I also vectorized the `deltav` function by rewriting it in terms of numpy calls. Previously it was using Pandas .apply for vectorization, but now it should be faster and simpler to use, working either with apply or simply being passed the entire dataframe.

Remaining issues:
- [x] Chained pn.bind issue? show_candidates only works the first time, to enable showing them; it doesn't let them be turned off afterwards. ETA: turns out to be an hv, not panel, issue.
- [ ] Update holoviews pin to get alpha and visible fixes once released, and re-run `anaconda-project lock`
- [x] Some of the cells could use more explanation -- what are we supposed to see? How do we interpret it?
- [x] I added a title to the final dashboard, but it should probably also have some explanatory text
- [x] Add a template?

### DIFF
--- a/exoplanets/anaconda-project-lock.yml
+++ b/exoplanets/anaconda-project-lock.yml
@@ -62,7 +62,7 @@ env_specs:
       - nest-asyncio=1.5.1=pyhd3eb1b0_0
       - olefile=0.46=py37_0
       - packaging=20.9=pyhd3eb1b0_0
-      - panel=0.11.3=pyhd3eb1b0_0
+      - panel=0.12.0=pyhd3eb1b0_0
       - param=1.10.1=pyhd3eb1b0_0
       - parso=0.8.2=pyhd3eb1b0_0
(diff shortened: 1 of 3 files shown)


philippjfr:
Can we wrap the dashboard in a template at the end? I don't think we should be adding further examples that don't use templates at this point.

jbednar:
Now requires fixes on holoviews master and holoviews branch `visible_update`; shouldn't be published until those changes are available.

PR #179, opened 2021-08-20
Explanation of earth's position in exoplanets example 

ablythed:
I added a few sentences clarifying the coordinate system as well as an arrow pointing to the earth in one of the plots.


jbednar (approved):
Looks great!

PR #180, opened 2021-08-23
Update holoviews version to 1.14.6a1

ablythed:


exoplanets/anaconda-project.yml
```diff
@@ -19,7 +19,7 @@ packages: &pkgs
   - panel
   - bokeh
   - hvplot
-  - holoviews
+  - holoviews>=1.14.6a1,<=1.14.6
```
jbednar, line 22: Oops! I know I suggested pinning <=1.14.6, but I forgot that any upper pin is taken care of by the lock file anyway. So I think we should only pin the lower bound, as a declaration of the minimum holoviews release required, and we would _expect_ higher holoviews releases to work as well in the future.

```suggestion
  - holoviews>=1.14.6a1
```

PR #186, opened 2021-08-28
Added accordion with background info to dashboard and fixed holoviews links

ablythed:


jlstevens:
Thanks!

Unfortunately I didn't notice this PR before rebuilding the project (and starting the live deployment). I'll rebase and do it here.

PR #187, opened 2021-09-21
Update gull tracking notebook

maximlt:
I've managed to run this notebook by fixing anaconda-project with this fix https://github.com/Anaconda-Platform/anaconda-project/pull/340. Except one tiny import change, it's running fine.

@jbednar does the lock file have to be updated and if so how?


jbednar:
Fix looks good, thanks!  

The locking instructions should be at https://examples.pyviz.org/make_project.html but I think they are missing; it should say that when first creating the project run `anaconda-project lock` and commit the resulting `anaconda-project-lock.yml` file, and when updating dependencies delete the local `envs` dir (if you've previously run the project locally) and the lock file, then run `anaconda-project lock` again. In this case I don't think any dependencies need to change, so the previous lock file should be fine.

PR #188, opened 2021-09-22
Website update

jlstevens:
In this PR I intend to build and deploy some of the new projects that have been recently added. There are many projects that need to be rebuilt and deployed and I will use this as an opportunity to figure out how best to automate things.


jlstevens:
Mystery solved!

In short, after fixing the syntax error in `datashader_dashboard`, I generated a new lock file which was working..for a few hours and long enough for the tests to go green. Unfortunately, it picked up a `libcurl` version that got pulled by conda-forge a few hours later, breaking the lock.

In short, I see two lessons to learn from this:

1. The step that parses all the project yamls should be robust to syntax errors and skip projects it cannot parse. We should have a mechanism to make sure all yamls are readable.
2. We have another reason to avoid conda-forge other than slow solve times: sometimes packages can be captured by lock files and pull off conda-forge breaking those locks.

jlstevens:
The exoplanets project is now live on the website together with the AE5 deployment.

As a lot of effort was put into fixing things, I will go ahead and merge this to master before opening a new PR to add/update the rest of the projects.

PR #192, opened 2021-11-09
Add the GrabCut example

maximlt:
This PR adds the GrabCut example that relies on the `earthsim` package, taken from [Earthsim's topics](https://earthsim.holoviz.org/topics/GrabCut.html). I've just added the header that's now common with all the examples and also a way to disable the warnings emitted by newer versions of `param` (using methods outside of the main `.param.` namespace). I couldn't install an older version of param due to some dependency conflict, and updating holoviews/geoviews (the causes of the warnings) to get a working environment seemed to tricky.

- [ ] Define maintainers / authors
- [ ] The lock file wouldn't solve on Windows so I just removed that platform, are we OK with that for this example?


maximlt:
Closing in favor of #193 .

The most important piece of this PR was the environment that allowed to run GrabCut **with** earthsim, which has been recorded in this issue: https://github.com/pyviz-topics/EarthSim/issues/327

PR #193, opened 2021-11-15
Add the GrabCut example

maximlt:
This PR adds the GrabCut example taken from [Earthsim's topics](https://earthsim.holoviz.org/topics/GrabCut.html). While the original example relied on the `earthsim` library, this example relies on the module `earthsim.py` that sits next to the notebook. This module has been extracted and adapted from `earthsim`. Doing so has allowed to significantly reduce the number of dependencies required to run the grabcut example, and as a consequence made it possible to update the remaining ones.

Compared to the original example:
* No more dependency on `quest` to download map tiles, instead the module relies on the function `get_tile_rgb` available in the `util.py` module of geoviews.
* The last step that used a `PolyAnnotator` to modify the contour generated during the main/previous stage has been entirely removed since annotators are currently broken in geoviews (https://github.com/holoviz/geoviews/issues/526).
* Added a few comments and some metadata (authors, last updated...).


maximlt:
@jlstevens if you don't mind having a look at this, I think that this time we're closer to what we wanted to have from the beginning! Too bad that the annotators seem to be broken in geoviews, I just realized it quite late in the process of updating the main grabcut class.

jbednar:
> annotators seem to be broken in geoviews

That's alarming!!

jlstevens:
Thanks for the PR!

I'll try this out tonight - this will be a nice test case for my deployment tool. :-)

> annotators seem to be broken in geoviews

That sounds like an issue to file on geoviews if you can submit a small, reproducible example...

maximlt:
@jlstevens thanks! And there was already an issue opened on geoviews 🙃

grabcut/grabcut.ipynb
```diff
@@ -0,0 +1,120 @@
...
+    "Last updated: November 14, 2021"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Extract a coast- and shoreline contour with GrabCut\n",
+    "\n",
+    "The GrabCut algorithm provides a way to annotate an image using polygons or lines to demark the foreground and background. The algorithm estimates the color distribution of the target object and that of the background using a Gaussian mixture model. This is used to construct a Markov random field over the pixel labels, with an energy function that prefers connected regions having the same label, and running a graph cut based optimization to infer their values. This procedure is repeated until convergence, resulting in an image mask denoting the foreground and background.\n",
```
jbednar, line 20:
```suggestion
    "The GrabCut algorithm provides a way to annotate an image using polygons or lines to demark the foreground and background. The algorithm estimates the color distribution of the target object and that of the background using a Gaussian mixture model. This is used to construct a Markov random field over the pixel labels, with an energy function that prefers connected regions having the same label, and running a graph-cut-based optimization to infer their values. This procedure is repeated until convergence, resulting in an image mask denoting the foreground and background.\n",
```

grabcut/grabcut.ipynb
```diff
@@ -0,0 +1,120 @@
...
+    "\n",
+    "In this example this algorithm is applied to map tiles to automatically extract a coast- and shoreline contour. First we specify a rectangular region to download the map tiles in using the `SelectRegionPanel`, then we can declare the ``GrabCutPanel`` to annotate the region and let the algorithm compute a contour."
```
jbednar, line 22:
```suggestion
    "In this example this algorithm is applied to map tiles to automatically extract a coast or shoreline contour. First we specify a rectangular region in which to download the map tiles using the `SelectRegionPanel`, then we can declare the ``GrabCutPanel`` to annotate the region and let the algorithm compute a contour."
```

grabcut/grabcut.ipynb
```diff
@@ -0,0 +1,120 @@
...
+    "])\n",
+    "dashboard = GrabCutPanel(select_region.get_image(), fg_data=[foreground], bg_data=[background], minimum_size=500, tolerance=0.001)\n",
+    "pn.Row(dashboard.param, dashboard.view())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Pressing the ``Update Contour`` runs the *GrabCut* algorithm on the selected region with the foreground and background polygons/paths given as input. The output of this process is a contour that delineates the foreground to the background, this countour is additionaly filtered and simplified. Try to set ``Minimum size`` to 1 and ``Tolerance`` to 200 to observe the effects of filtering and simplifying the contour, respectively. To speed up the calculation we can also downsample the image before applying the Grabcut algorithm by reducing the value of ``magnification``.\n",
```
jbednar, line 80:
```suggestion
    "Pressing the ``Update Contour`` runs the *GrabCut* algorithm on the selected region with the foreground and background polygons/paths given as input. The output of this process is a contour that delineates the foreground and background. This countour is also filtered and simplified to make it smaller to export and easier to edit. Try to set ``Minimum size`` to 1 and ``Tolerance`` to 200 to observe the effects of filtering and simplifying the contour, respectively. To speed up the calculation we can also downsample the image before applying the Grabcut algorithm by reducing the value of ``magnification``.\n",
```

grabcut/grabcut.py
```diff
@@ -0,0 +1,472 @@
+"""
+Module that supports the grabcut.ipynb example.
+
+Its content has been extracted and adapted from earthsim.
```
jbednar, line 4:
```suggestion
Its content has been extracted and adapted from earthsim.holoviz.org.
```

jbednar (approved):
Nice!

grabcut/grabcut.ipynb
```diff
@@ -0,0 +1,120 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Grabcut\n",
+    "\n",
+    "Written by Philipp Rudiger<br>\n",
+    "Created: November 14, 2021<br>\n",
```
maximlt, line 10:
```suggestion
    "Created: November 16, 2018<br>\n",
```

Alright I'm going to set it to the date when this PR (https://github.com/pyviz-topics/EarthSim/pull/253) was merged in earthsim, some time in November 2018.

jlstevens:
I've rebased this PR and if I can get it to build I'll merge to master and update the public website.

jlstevens:
Maxime just reminded me that there is one last change to be made to the grabcut project: we need to serve a panel dashboard with `.servable()` if we want to deploy (due to the lack of openGL in the deployment containers).

maximlt:
@jlstevens I've made Grabcut servable by creating a small dashboard with a pipeline that includes the first two stages introduced in the notebook. I've not added the 3rd stage (allow to manually edit the contour obtained after running grabcut) since the 4th stage which should allow to download the output (or do something with it, anything really) isn't included in the notebook, since it wasn't included in Earthsim. I don't even know if it's possible to export a Path or Polygon to a shapefile with geoviews?

Last section of the notebook:
image

First stage of the deployed app:
image

jlstevens:
Looks great!

A two stage pipeline is fine for now: I'll see if I can deploy a running version of this project shortly and if that works well, we can think about whether it is worth introducing more stages.

jbednar:
Cool! I believe shapely can export a shapefile.

maximlt:
> Cool! I believe shapely can export a shapefile.

That would be with `fiona`, I don't think `shapely` does any I/O. Maybe also with `pyshp`.

ahuang11:
Oh man, I can definitely understand why Maxime had concerns about panel-chat-examples and future maintenance; for old, unmaintained repos, its a nightmare.

Took me the whole day to get it barely functioning. It's cool when it works, but super, extremely finicky.

https://github.com/holoviz-topics/examples/assets/15331990/49da16d0-c354-4ac7-aa65-da9226e0041e

This drained the life out of me, and I'd prefer to rebuild existing ideas thru slightly more modern tools like https://twitter.com/giswqs/status/1649416858645282822 instead next time.

grabcut/grabcut.py
```diff
@@ -320,8 +320,8 @@ def view(self):
         dmap = hv.DynamicMap(self.extract_foreground)
         dmap = hv.util.Dynamic(dmap, operation=self._filter_contours)
         dmap = hv.util.Dynamic(dmap, operation=self._simplify_contours)
-        return (regrid(self.image).options(**options) * self.bg_paths * self.fg_paths +
```
ahuang11: Why was the regrid add here? It kept making the shapes mismatched and took me forever to find it :(

grabcut/grabcut.py
```diff
@@ -320,8 +325,10 @@ def view(self):
         dmap = hv.DynamicMap(self.extract_foreground)
         dmap = hv.util.Dynamic(dmap, operation=self._filter_contours)
         dmap = hv.util.Dynamic(dmap, operation=self._simplify_contours)
-        return (self.image.opts(**options) * self.bg_paths * self.fg_paths +
-                dmap.opts(**options))
+        return pn.Row(
```
ahuang11, line 328: If two freehanddraw tools collide, https://github.com/holoviz/holoviews/issues/6140

`merge_tools` also broken :'( https://github.com/holoviz/holoviews/issues/6139

grabcut/grabcut.py
jlstevens: Presumably to make it possible to display *really* large images without causing issues in the browser? That would be my initial guess at least.

maximlt:
> Oh man, I can definitely understand why Maxime had concerns about panel-chat-examples and future maintenance; for old, unmaintained repos, its a nightmare.

Welcome to the gang :D Code that doesn't run dies quicker than one would think!

grabcut/grabcut.py
ahuang11: Simon mentioned using different tooltip names and it worked.
https://github.com/holoviz/holoviews/issues/6140

ahuang11:
Would like a review on this.

grabcut/grabcut.ipynb
```diff
@@ -0,0 +1,149 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Grabcut\n",
+    "\n",
+    "Written by Philipp Rudiger and Maxime Liquet<br>\n",
+    "Created: November 16, 2018<br>\n",
+    "Last updated: January 7, 2021"
```
jbednar, line 11: Do the dates auto-update on merge?

jbednar (approved):
Looks great to me! No longer any outlandish imports!

grabcut/grabcut.ipynb
ahuang11: No I don't think so.

ahuang11, line 11:
```suggestion
    "Last updated: April 26, 2024"
```

ahuang11:
Can we merge this?

jbednar:
Please!

maximlt:
Let me review please, I see some changes to make, to the project config for instance...

maximlt:
@ahuang11 did you use anaconda-project to set up your environment to work on Grabcut? I see the project is not locked on `osx-arm64`.

ahuang11:
It's been a while so I don't really remember. I believe I did, but I didn't lock the project.

maximlt:
> Looks great to me! No longer any outlandish imports!

Sorry to come back to this old message @jbednar, I'm sure @ahuang11 would agree that "great" isn't an adjective that qualifies GrabCut 🙃 I don't know if you ran the notebook but I just did, after relocking since the last lock didn't support osx-arm64, and, oh boy, that's a painful example to run through. I wrote down a list of bugs to resolve and suggestions to improve the UX, which is pretty terrible at the moment. It's a non-trivial amount of work and merging GrabCut is just a nice-to-have on my current Examples todo list (https://github.com/holoviz-topics/examples/issues/383), so I won't prioritize it, unless told so.

- [ ] It's very easy to get confused with mapping background/foreground to water/land (or the other way around!)
- [ ] Display activity (e.g. loading indicator, notifications)
- [ ] Fix the update contour algorithm that somehow seems to produce two contours instead of one?
image
- [ ] Does the PolyEdit tool still work? I've tried many, many, times to edit the contour(s) in the notebook and never managed to do so. Among many tries, I've managed to edit a polygon on the HoloViews reference page only once, enough to show me it somehow works, which annoyed me quite decently as I never managed to get it to work again (https://holoviews.org/reference/streams/bokeh/PolyEdit.html)
- [ ] Since the update contour algorithm is broken, the paths to polys procedure produces bad results
image
image

versus (from https://earthsim.holoviz.org/topics/GrabCut.html)
image
- [ ] On the final dashboard, I managed to pass the first stage, though creating the box only worked when I used SHIFT+drag (double click did nothing). I didn't manage to get through the second stage, I didn't try to debug it.

In my OP 3 years ago I wrote:

> The last step that used a PolyAnnotator to modify the contour generated during the main/previous stage has been entirely removed since annotators are currently broken in geoviews (annotators wont work in geoviews  holoviz/geoviews#526).

[ ] That issue has been fixed so the annotator little demo could be reintroduced, in theory at least (see again https://earthsim.holoviz.org/topics/GrabCut.html)

jbednar:
> > Looks great to me! No longer any outlandish imports!

> Sorry to come back to this old message @jbednar, I'm sure @ahuang11 would agree that "great" isn't an adjective that qualifies GrabCut 🙃 I don't know if you ran the notebook but I just did, after relocking

What looked great to me (and still looks great) is that anaconda-project.yml no longer includes anything apparently crazy; those are all released, normal packages, and that's a great improvement.  I don't know if I ran the code at the time, and I haven't run it today, but I agree that the behavior you outline above doesn't seem good.  This example was one of the motivators for how we developed our ideas around editing and annotating shapes, so it's not surprising that it has a lot of barely ok ways of doing those things.  I have very often wanted to demo the ideas covered by this example, but have been stymied by it being unable to build or having functionality that wasn't working. I'm perfectly happy with a different approach to showing the same thing, and don't need the algorithm involved to be grabcut specifically, though it's nice that it's simpler than e.g. Segment Anything. In any case the actual segmentation performance isn't the main point of the example, but instead it's about being able to draw shapes, then select regions by value based on the drawings, then annotate them, and then use that shape in subsequent computations. Those are all _very_ important capabilities that we are currently almost entirely hiding from our users without having this example running.

maximlt:
Oh yes, I agree the concept overall is okay and this example is a good candidate for the gallery. It's just that the example needs to be fixed (and maybe some upstream libraries too) before we publish it. I'm a bit sad we can't, as all the work I did on this repo was originally motivated by this example, merging new examples 3 years ago was pretty brittle. I don't have the time to work on these fixes at the moment, maybe early next year. Or someone else could pick this up. I don't think working on another non-GrabCut example using the same functionality will be less work so let's fix this one.

Another suggestion to improve the example:

- [ ] The main and most interesting code is in the `grabcut.py` module. It should instead be exposed directly in the notebook, there's no need for this module to exist (remember this example was extracted from earthsim.pyviz.org that was itself a package).

maximlt:
>  Does the PolyEdit tool still work? I've tried many, many, times to edit the contour(s) in the notebook and never managed to do so. Among many tries, I've managed to edit a polygon on the HoloViews reference page only once, enough to show me it somehow works, which annoyed me quite decently as I never managed to get it to work again (https://holoviews.org/reference/streams/bokeh/PolyEdit.html)

Ok, so we figured this out! Bokeh's behavior changed in version 3.4, to enter the edit mode with the PolyEditTool you no longer double-click but press tap (more than 300ms). This was changed in https://github.com/bokeh/bokeh/pull/12831. I've opened an issue to track updating HoloViews' docs accordingly https://github.com/holoviz/holoviews/issues/6474

PR #194, opened 2021-12-06
Added project thumbnails and fixing warnings

jlstevens:
This PR will add explicit thumbnails to all projects so we can ditch the old thumbnailing code that causes conda solve issues.

In addition I will address some of the simpler issues with these projects (but not everything, especially issues relating to environment definitions and locks).


jlstevens:
@jbednar That handles all the issues I can detect other than missing and stale locks and incorrectly defined package specs.

I'm not expecting the tests to pass: each project will need to be updated individually (i.e. a doc build) anyway. Please go ahead and merge if the changes look good to you!

jbednar (approved):
Looks good, thanks!

PR #195, opened 2022-01-19
Run archive_projects on doc build

philippjfr:
This wasn't being run presumably resulting in the archives never getting updated.


jbednar:
That's an odd thing to have missed. Thanks! Any idea if the test failure needs to be addressed here? Otherwise happy to merge.

maximlt:
The docs build has a step that creates the assets, and it does pull the project archives from the evaluated branch (in doc/project_name/projec_name.zip) to the assets (in assets/). This wasn't missed :)

PR #196, opened 2022-01-26
Attempting to fix website

jlstevens:
Reverting the last commit to see if that fixes things.


jlstevens:
I'll merge now so the website doesn't get in the broken state. Problems left to fix:

- [ ] Tests failing due to libtiff?
- [ ] I'll reopen the PR with the fixes I made that have been reverted. I'll have to find out why fixing errors in projects breaks the websites (I suspect it is due to state that holds onto incorrect project names somewhere).

PR #197, opened 2022-01-27
Reintroducing fixes from #194

jlstevens:
Reapplies fixes from #194 which were later reverted to fix the website.

*BEWARE*: Please be careful when working on this PR to avoid leaving https://pyviz-dev.github.io/examples/ in a broken state as this will propagate to the public website if anyone cuts a new website release!

I'm not sure why the fixes here would break the thumbnails and doc building but I suspect old (incorrect) project names or thumbnail file names are expected and sticking around somewhere...


Azaya89:
@maximlt Is this PR still relevant considering the latest updates?
cc: @jlstevens

jlstevens:
I suspect this PR may indeed be out-of-date, especially now the website has been updated many times and the thumbnails look good. However, there are also improvements to filenames that I see in the project YAMLs that I would double check have been applied everywhere before closing this PR.

PR #199, opened 2022-04-21
Updated nyc_buildings

jbednar:
Updated nyc_buildings to use higher res, have a dashboard, and explain more. At least on my Mac Retina display, the default PlotSize.scale made the output much blockier than it needed to be. The datashader-rendered polygons are now just as sharp as the Bokeh-rendered overlay:

image

Issues:
- [ ] I added a non-categorical plot at the start, and it seems a _lot_ faster to render than the categorical plot. Is that expected to be unavoidable?
- [ ] I want the dashboard to be responsive to the browser tab screen size, but I have no idea how to convince it to do that. If you know, please fix this and update https://github.com/holoviz/holoviews/pull/5278 .

### DIFF
--- a/nyc_buildings/anaconda-project.yml
+++ b/nyc_buildings/anaconda-project.yml
@@ -27,6 +27,9 @@ packages: &pkgs
 dependencies: *pkgs
 
 commands:
+  dashboard:
+    unix: panel serve nyc_buildings.ipynb --show
+    supports_http_options: true
   notebook:
     notebook: nyc_buildings.ipynb
   test:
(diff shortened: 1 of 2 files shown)


jbednar:
For Polygon, TriMesh, Area, Raster, and QuadMesh, I agree. Because those plot types  have a spatial extent, a finer grid for sampling them is always a win (at some cost in speed and document size).

However, for Points and non-antialiased lines, a finer resolution can make them much more difficult to see. The results when using points spreading and line spreading or antialiasing can look great with the increased resolution, but without one of those two mechanisms for making the item have a spatial extent, the additional resolution can be a big problem.  So I'm not sure if we can move to a higher default PlotSize scale unless we also update everything to have a spatial extent.

Azaya89:
This example is being worked on here #386

maximlt:
Ok thanks @Azaya89, please come back and comment on this PR when you have made more progress on https://github.com/holoviz-topics/examples/pull/386. Make sure you have a look at the content of this PR, and the comments from Jim, to see what you can incorporate/improve/answer to.

Azaya89:
I've incorporated most of the changes here into #386, although the PR is not finished yet.

PR #206, opened 2022-08-08
Fixed handling of defaults and nodefaults in channel lists

jlstevens:
Fix for the issue described in  https://github.com/Anaconda-Platform/anaconda-project/pull/387 and https://github.com/Anaconda-Platform/anaconda-project/issues/388.

In short:

* The `defaults` channel was implicitly picked up from the `.condarc`
* `nodefaults` was ignored by anaconda-project (though it is respected by conda itself)
* The latest anaconda-project version now (correctly) ignores the `~/.condarc` which has implications for backwards compatibility.

This PR attempts to fix this issue as follows:

* Make `defaults` explicit everywhere by appending it to the channel list if it is not stated. This should be backwards and forwards compatible with anaconda-project.
* Remove any instances of` nodefaults`, making sure `defaults` appears at the end of the list and noting that the *intention* was to avoid using defaults in a comment.

I *believe* this should fix compatibility but this PR has the possibility of breaking lots of things so please make sure you agree with this strategy @jbednar @maximlt !

### DIFF
--- a/attractors/anaconda-project.yml
+++ b/attractors/anaconda-project.yml
@@ -9,6 +9,7 @@ labels:
 
 channels:
 - pyviz
+- defaults
 
 user_fields: [labels, skip, maintainers]
 
--- a/bay_trimesh/anaconda-project.yml
+++ b/bay_trimesh/anaconda-project.yml
@@ -9,7 +9,7 @@ labels:
 
 user_fields: [labels, skip, maintainers]
 
-channels: []
+channels: [defaults]
 
 packages: &pkgs
 - python=3.7
--- a/boids/anaconda-project.yml
+++ b/boids/anaconda-project.yml
@@ -9,7 +9,7 @@ labels:
 
 user_fields: [labels, skip, maintainers]
 
-channels: []
+channels: [defaults]
 
 packages: &pkgs
   - python=3.7
--- a/carbon_flux/anaconda-project.yml
+++ b/carbon_flux/anaconda-project.yml
@@ -11,7 +11,7 @@ labels:
 
 user_fields: [labels, skip, maintainers]
 
-channels: [pyviz]
+channels: [pyviz, defaults]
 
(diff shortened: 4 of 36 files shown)


jlstevens:
Note: A separate PR should go through the projects trying to use `nodefaults` with the dev version of anaconda-project that actually respects it, to see if those projects *actually* function without defaults.

Also, I believe these changes shouldn't require relocking everything but I will have to double check.

jlstevens:
Tests currently failing due to carbon flux not having test data. I'll have a look into this shortly.

maximlt:
I totally approve this approach.

Note that you won't be able to build many examples, even before anaconda-project 0.11 they were mostly all broken. I have to resume the work on fixing that.

> Tests currently failing due to carbon flux not having test data. I'll have a look into this shortly.

That sounds strange. For reference here's a table that sums up what data the projects need/use: https://github.com/pyviz-topics/examples/issues/202#issuecomment-1166689039

Also and seems to be of importance, `anaconda-project` is pinned to an older version in the `ecosystem_setup` command:

https://github.com/pyviz-topics/examples/blob/10772f10bc5e8dfd7c3bd8a894d372135176b39a/dodo.py#L17-L23

jlstevens:
> ... Also and seems to be of importance, anaconda-project is pinned to an older version in the ecosystem_setup command

Well spotted! 

That said, this pin is for the infrastructure here, this PR as a whole is for users who may install any version of anaconda-project they want to try running these projects locally. Knowing the anaconda-project version is useful though, to debug any issues cropping up due to channel behavior.

maximlt:
> Also, I believe these changes shouldn't require relocking everything but I will have to double check.

I believe projects don't need their lock to be updated to run, but I also believe users will get a warning. I think for now it's fine and will go ahead and merge. Thanks Jean-Luc!

PR #208, opened 2022-08-29
Fix link in attractors

### LINKED ISSUE #207: Links on Attractors panel not found
The files on https://examples.pyviz.org/attractors/attractors_panel.html redirect to page not found

image

Hoxbro:
Fixes #207

### DIFF
--- a/attractors/attractors_panel.ipynb
+++ b/attractors/attractors_panel.ipynb
@@ -29,7 +29,7 @@
     "\n",
     "## Attractor definitions\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jbednar (approved):
Yeah; the existing link works when it's a notebook, but not on the website. I guess the ideal would be to expand such links to refer to github automatically when the website is built, but that seems painful to implement. The new link will fail for offline docs exploration, so it's not ideal, and this seems like an improvement in the meantime.

maximlt:
Fixed in https://github.com/holoviz-topics/examples/pull/290

PR #211, opened 2022-12-08
Updated dependencies and lock

jlstevens:
Update since the current lock file failed. Dependencies have updated, only new issue is the weird legend:

image


jlstevens:
This is the version that is now deployed here: https://ship-traffic.pyviz.demo.anaconda.com/ship_traffic

Merging.

ship_traffic/anaconda-project.yml
```diff
@@ -10,27 +10,23 @@ labels:
 user_fields: [labels, skip, maintainers, user_fields]
 
 channels:
-- pyviz
-- conda-forge
+- pyviz/label/dev
```
jbednar, line 13: What's label/dev important for? I don't see any dev releases below.

jlstevens: Good catch. I can try the solve with just `pyviz` which should work..then I can update the lock again. The original project yaml before I unpinned used a number of alphas.

PR #250, opened 2023-01-24
Rebuild attractors

maximlt:


maximlt:
Fixed two bugs in `attractors_panel` which together caused to app to be permanently down:

- the app allowed the user to overwrite `attractors.yml` from which the set of examples is loaded
- one of these examples was invalid, and when a user would shuffle/sort the examples AND save then this invalid example would end up being first in the list of examples for that attractor type, the app gets initialized by the first example for each attractor and would then fail, on start.

The changes made were:
- prevented users from overwriting the default file, now they save their changes to another file that cannot be named as the default file
- removed the invalid attractor example

I made a couple of small changes, for instance shuffling/sorting was not modifying in place the list of examples and as such the scrubber options would not get updated.

PR #254, opened 2023-01-31
Add stable-diffusion example using panel

sandhujasmine:


stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+
+# optional environment variables
+variables: {}
+
+# optional downloads
+downloads: {}
+platforms:
+- linux-64
+- osx-64
+- win-64
```
sandhujasmine, line 120: Only tested on linux w/ GPU

stable-diffusion/.projectignore
```diff
@@ -0,0 +1,25 @@
+# MOST projects DO NOT NEED this file. Only those which publish a data/ folder
```
maximlt, line 1: Most projects don't need to submit their `.projectignore` file, I think this will apply to this project too.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
+# To reproduce: install 'anaconda-project', then 'anaconda-project run'
+
+# required: name MUST match the directory name.
+#           name must be of the form r'[a-z_]+'
+name: stable-diffusion
```
maximlt, line 5: The historic pattern on the project names is to use `_` instead of `-`. You will have to update this name, the directory name, the notebook name and the thumbnail name.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+  created: 2022-01-30
+  # List of the GITHUB handle of the authors/maintainers
+  maintainers:
+  - "sandhujasmine"
+  # List of labels displayed in the project card
+  # Each label must be a name (e.g. panel) that
+  # refers to a SVG badge located in doc/_static/labels
+  # (e.g. doc/_static/labels/panel.svg)
+  labels:
+  - "channel_conda-forge"
```
maximlt, line 22: We'll remove the channel labels, listing `panel` only should be enough.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+  - "panel"
+
+  ### OPTIONAL ###
+  # Listed deployments will by default automatically be started.
+  # Maximum number of deployments is 2.
+  # Each deployment must declare the command it deploys, options
+  # include "notebook" or "dashboard".
+  deployments:
```
maximlt, line 30: I assume this project will have no deployment on AE5, you can remove all of that.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+  no_data_ingestion: false
+
+# required: (needed internally)
+user_fields: [examples_config]
+
+# required: list of the channels needed to solve the environment
+channels:
+- pyviz
+- defaults
+- conda-forge
```
maximlt, line 57: I would guess that depending on `conda-forge` is going to be enough for this project. There are probably some missing packages on `defaults` anyway, and both `conda-forge` and `pyviz` have the latest version of Panel.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+    # the `--rest-session-info --session-history -1` options are required
+    unix: panel serve --rest-session-info --session-history -1 stable-diffusion.ipynb
+    supports_http_options: true
+  # required: a project must have a test command that references the 'test' env_spec
+  test:
+    unix: pytest --nbsmoke-run -k *.ipynb --ignore envs
+    windows: pytest --nbsmoke-run -k *.ipynb --ignore envs
+    env_spec: test
+  # required: a project must have a lint command that references the 'test' env_spec
+  lint:
```
maximlt, line 97: I'm working on a branch that will remove the need for each project to define the `lint` command, you can remove it already.

stable-diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,120 @@
...
+    notebook: stable-diffusion.ipynb
+    # shortcut to open all the notebook
+    # notebook: .
+  # if you intend to deploy a panel/lumen app, set the `dashboard` command
+  dashboard:
+    # the `--rest-session-info --session-history -1` options are required
+    unix: panel serve --rest-session-info --session-history -1 stable-diffusion.ipynb
+    supports_http_options: true
+  # required: a project must have a test command that references the 'test' env_spec
+  test:
```
maximlt, line 92: Do you think it's going to be possible to run this notebook on the CI?

stable-diffusion/anaconda-project.yml
maximlt: Just FYI. Pretty sure most of the examples here have not been tested on all platforms. I'd guess that they were built on a Mac and then somewhat tested on the CI on Linux. The GPU is definitely an issue for the CI.

stable-diffusion/stable-diffusion.ipynb
```diff
@@ -0,0 +1,437 @@
...
+   "cell_type": "markdown",
+   "id": "0c10ed41-57e2-4e7f-a34d-16c94dccbb67",
+   "metadata": {},
+   "source": [
+    "# GUI for Stable Diffusion\n",
+    "\n",
+    "# Stable Diffusion\n",
+    "Written by Jasmine Sandhu (Acknowledgements: Jim Bednar, Maxime Liquet, Philipp Rudiger)<br>\n",
+    "Created: Jan, 2023<br>\n",
+    "Last updated: Jan, 2023"
```
maximlt, line 13: You no longer need to include this metadata header in the notebook, this will be included automatically during the doc build, pulling information from the `anaconda-project.yml` file.

maximlt (commented):
I've added some comments, before I can point you to the updated docs :)

stable-diffusion/anaconda-project.yml
sandhujasmine: Yep - saw the updated docs. That's usually best practice for *nix - not sure why/how I missed that. I'll fix. Thanks

stable-diffusion/anaconda-project.yml
sandhujasmine: To deploy on AE5, we would need to add a GPU node with a resource profile and point this deployment to use that resource profile. 

I'll remove this section for now but adding this to the TODO list above. It would be nice.

stable-diffusion/anaconda-project.yml
sandhujasmine: For dev setup I didn't use defaults. 
I tested this using these channels since I thought we may want to point to main. But ok to remove.

stable-diffusion/anaconda-project.yml
sandhujasmine: The deployment will likely fail. But it can still validate the `anaconda-project` file.
Currently it is [failing when installing nbval](https://github.com/pyviz-topics/examples/actions/runs/4056819872/jobs/6981799733)

I don't believe tests will work in current state, since it looks like nbval is used to run the notebook cells which fail on CPU machines. I'll add a task to update this to work on CPU if GPU is not found - I think that is useful in any case.

stable_diffusion/stable_diffusion.ipynb
```diff
@@ -167,7 +170,6 @@
     "            except OSError:\n",
     "                pipelines[m] = init_model(m, torch_dtype=torch.float16, local_files_only=False)\n",
     "            \n",
-    "        \n",
     "    if torch.cuda.is_available():\n",
     "        pseudo_rand_gen = torch.Generator(device='cuda')\n",
```
sandhujasmine, line 174: Use cuda only if available. This should allow the cells to run on a cpu machine.

stable_diffusion/anaconda-project.yml
```diff
@@ -0,0 +1,80 @@
...
+    supports_http_options: true
+  # required: a project must have a test command that references the 'test' env_spec
+  
+# optional environment variables
+variables: {}
+
+# optional downloads
+downloads: {}
+platforms:
+- linux-64
```
sandhujasmine, line 80: limited to linux-64 for now.

stable-diffusion/anaconda-project.yml
sandhujasmine: Removed osx-64 and win-64 for now.

stable-diffusion/.projectignore
sandhujasmine: Removed!

stable-diffusion/anaconda-project.yml
sandhujasmine: Removed this section

stable-diffusion/anaconda-project.yml
sandhujasmine: Added some checks to run the cells on CPU. The cells should run with nbval

stable_diffusion/anaconda-project.yml
```diff
@@ -87,5 +79,30 @@ variables: {}
...
+  stable-diffusion-m1:
+    description: Env for osx-arm64 M1 for running app
+    packages:
+    - python>=3.10,<3.11.0a0
+    - pytorch >=1.13.1
+    channels:
+    - pytorch
+    - conda-forge
+    platforms: 
+    - osx-arm64
```
sandhujasmine, line 109: Define an `env_spec` to install custom packages like pytorch for M1

stable-diffusion/anaconda-project.yml
sandhujasmine: Using `env-spec` to build custom env for M1, see comment here: https://github.com/pyviz-topics/examples/pull/254/commits/7d30c94c4bab14de000b87f7747eb67a6a94e1c6#r1119620931

stable_diffusion/stable_diffusion.ipynb
```diff
@@ -218,45 +221,46 @@
...
     "    seed = param.Integer(label='Random seed', default=1,\n",
     "                         bounds=random_int_range, step=10, precedence=1, doc=\"\"\"\n",
     "        Seed controlling the noise values generated.\"\"\")\n",
-    " \n",
+    "\n",
     "    generate = param.Event(precedence=1)\n",
-    "    \n",
-    "    param.depends(\"generate\")\n",
+    "\n",
+    "    @param.depends(\"generate\")\n",
```
maximlt, line 253: @sandhujasmine it seems that there was a missing `@` on this line, which I believe meant that the decorator wasn't applied. We'd better double check what's the correct change, either what I've done or removing entirely this line.

jbednar: That's odd; seems like it wouldn't have worked without that!

maximlt:
@sandhujasmine I did a few changes on the branch, mostly to get the project validated and to get it to be tested and built using the macos runner, that has more memory than the ubuntu ones which were failing because of that.

PR #256, opened 2023-02-01
Move the gallery extension to this repo

maximlt:
The goal with this PR is to move the gallery extension to this repo. I realized that the code dedicated to build the gallery here was pretty self-contained, had already quite diverged from the other more doc oriented galleries, and was meant to diverge even more (include link to deployments under the thumbnails, sharing on Twitter/LinkedIn, categories, etc.).

ref: https://github.com/pyviz-dev/nbsite/pull/256

I think that eventually the dependence of `nbsite` could be entirely removed to build this site.


philippjfr:
Does that mean we can simplify the nbsite implementation?

maximlt:
> Does that mean we can simplify the nbsite implementation?

Yes, not as much as one would hope though. I think that will be limited to removing the "inlining" code path, that was anyway broken.

Something this extension no longer relies on is the `deployment_url` template, as the project are now declaring their deployments and the system is in charge of deploying them and building the docs linking them. Panel seems to be using `deployment_url`, if at some point it no longer relies on that then that could also be removed.

PR #265, opened 2023-04-28
Updated census notebook and environment

jbednar:
Updated the census notebook as much as seems feasible right now. No longer needs geoviews, since holoviews has long included tile support and the data has already been projected to web mercator anyway. Also now uses `rasterize`, Bokeh-based colorbars, and hovering in the final example, along with more recent tools in the environment. There are several remaining issues that don't block merging but are important to fix:

- [ ] The environment pins to dask=2022.11.1 because later versions of dask hit [#1206](https://github.com/holoviz/datashader/issues/1206)
- [ ] pandas is pinned <2 because the earlier versions of dask don't support pandas 2
- [ ] There's a (hopefully temporary) workround `df = df.categorize('race')` in there to avoid hitting [#1202](https://github.com/holoviz/datashader/issues/1202)
- [ ] The parquet reader is set to `fastparquet` due to issues with the default `pyarrow` engine, though [#1202](https://github.com/holoviz/datashader/issues/1202)'s fix may also fix that issue
- [ ] The conda channel is set to conda-forge rather than defaults, because defaults has only an old and incompatible version of `fastparquet` (0.5.0, IIRC).
- [ ] The environment file doesn't enable `arm-osx64` for locking, because the environment fails to solve on that platform

Once these other issues are addressed it would be nice to update the census project again, removing the pins and workarounds.


census/anaconda-project.yml
```diff
@@ -53,14 +42,9 @@ downloads:
...
-      - pytest=4.4.1
-    dependencies: *testpkgs
-    
+
 platforms:
 - linux-64
 - osx-64
 - win-64
-
+# - arm-osx64 # unable to lock
```
maximlt, line 50: Easy mistake :)

```suggestion
- osx-arm64
```

census/anaconda-project.yml
```diff
@@ -1,45 +1,34 @@
...
   maintainers:
   - jbednar
   labels:
   - datashader
-  - geoviews
 
 user_fields: [examples_config]
 
-channels: [pyviz,conda-forge,defaults]
+channels: [conda-forge]
```
maximlt, line 13: When conda-forge is used, I suggest adding `nodefaults` to let anaconda-project know that it should exclude `defaults`:

```suggestion
channels: [conda-forge, nodefaults]
```

maximlt (commented):
I made a couple of suggestions that would impact the lock, I can relock if you want.

EDIT: Relocked after applying the suggestions.

PR #271, opened 2023-05-26
Update census example to use datashader 0.14.5a1

ianthomas23:
Work in progress to update the census example environment to use the latest `datashader`, `bokeh`, `panel`, etc. Initially using `datashader` dev release `0.14.5a` from `pyviz/label/dev` channel, and other packages from `conda-forge` rather than `defaults` until everything required is on `defaults`, which is well underway. The other change is to use `pyarrow` rather than `fastparquet`.

It results in the following packages used:

| Package | Before this PR | With this PR |
| --- | --- | --- |
| bokeh | 2.4.3 | 3.1.1 |
| dask | 2022.11.1 | 2023.5.0 |
| datashader | 0.14.4 | 0.14.5a1 |
| holoviews | 1.16.0 | 1.16.0 |
| pandas | 1.5.3 | 2.0.1 |
| panel | 0.14.4 | 1.0.3 |
| pyarrow | N/A | 12.0 |

It still includes the workaround for https://github.com/holoviz/datashader/issues/1202.

Tested locally on `osx-arm64`, historically one of the most problematic. It gives a bunch of `datashader` warnings, all of which are fixed and will be in the imminent `0.15.0` release. The output renders correctly for me except one plot is missing with a `tornado` error reported at the command-line, and the interactive plots at the end work correctly.

Given these results, I conclude that there is nothing fundamentally wrong with `datashader` to prevent going ahead with the `0.15.0` release. Once that is done here is a possible plan:

1. Update environment to use `datashader 0.15.0` to confirm that warnings have gone.
2. Ensure all `holoviz` packages are available on `defaults`.
3. Update environment to use `defaults` rather than `conda-forge` and `pyviz/label/dev`.
4. Investigate and/or deal with the `tornado` error if it still occurs.
5. Deal with https://github.com/holoviz/datashader/issues/1202.

### DIFF
--- a/census/anaconda-project.yml
+++ b/census/anaconda-project.yml
@@ -10,7 +10,8 @@ examples_config:
 
 user_fields: [examples_config]
 
-channels: 
+channels:
+- pyviz/label/dev  # Temporary until datashader 0.15.0 is on conda-forge/defaults
 - conda-forge
 - nodefaults
 
@@ -18,13 +19,11 @@ packages: &pkgs
 - python=3.10
 - notebook
 - colorcet
-- dask=2022.11.1
-- datashader
-- fastparquet
+- dask
+- datashader=0.14.5a1
 - holoviews
-- python-snappy
 - param
-- pandas<2
+- pandas
 
 dependencies: *pkgs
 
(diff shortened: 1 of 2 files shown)


Azaya89:
#430 will (has already) address(ed) this update so I'm closing this in favor of that one.

PR #273, opened 2023-05-29
Update landsat example

sandhujasmine:
- [x] remove pins and create new lock file. Add arm-x64 for OSX. Tested on Apple M1 Pro.
- [x] replace deprecated xarray.open_rasterio with rioxarray.open_rasterio 
- [x] ~pin tornado. I've been having rendering issues on OSX and seeing the following zmq error on OSX~ It is likely sufficient to pin `jupyter_client` [See hv-5699](https://github.com/holoviz/holoviews/issues/5699#issuecomment-1569025301)

<details>
<summary>

#### Exception which appears to be resolved with pinning tornado to older version

> Update: Replace the stacktrace below with the correct one

I've been experiencing rendering issues, along with a JS syntax error. As per this [jupyterhub discourse](https://discourse.jupyter.org/t/jupyter-notebook-zmq-message-arrived-on-closed-channel-error/17869/2) with similar error, the proposed solution of pinning tornado server seems to be consistently working correctly. The error I see (see below) points to the asyncio event loop so pinning tornado to older version does seem to point in the right direction. 

</summary>

```
[I 17:27:28.645 NotebookApp] Replaying 2 buffered messages
OMP: Info #276: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
[E 17:29:18.786 NotebookApp] Uncaught exception, closing connection.
    Traceback (most recent call last):
      File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 695, in _handle_events
        self._handle_write()
      File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 965, in _handle_write
        self._write_buffer.advance(num_bytes)
      File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 182, in advance
        assert 0 < size <= self._size
               ^^^^^^^^^^^^^^^^^^^^^^
    AssertionError
Exception in callback None()
handle: <Handle cancelled>
Traceback (most recent call last):
  File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/platform/asyncio.py", line 192, in _handle_events
    handler_func(fileobj, events)
  File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 695, in _handle_events
    self._handle_write()
  File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 965, in _handle_write
    self._write_buffer.advance(num_bytes)
  File "/Users/jasmine/projects/examples/landsat/envs/default/lib/python3.11/site-packages/tornado/iostream.py", line 182, in advance
    assert 0 < size <= self._size
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

</details>


landsat/anaconda-project.yml
```diff
@@ -10,38 +10,30 @@ examples_config:
 
 user_fields: [examples_config]
 
-channels: [pyviz, defaults]
+channels: [pyviz, defaults, conda-forge]
 
 packages: &pkgs
-- python=3.7
+- python
```
maximlt, line 16: I think our convention is to pin `python` in the project file

(There's a discussion somewhere on the GH repo where we even suggest that the convention could be to add a `>=` pin for all the packages - well except Python I guess - to track what minimum versions of each direct dependency were required to run the project vs. the dependencies that end up being locked up. As there's nothing to automate that process and it's kind of a pain to do manually since you then need to re-lock as anaconda-project finds out the project file changed, I doubt a little we'll manage to enforce that)

landsat/anaconda-project.yml
```diff
@@ -10,38 +10,30 @@ examples_config:
 
 user_fields: [examples_config]
 
-channels: [pyviz, defaults]
+channels: [pyviz, defaults, conda-forge]
```
maximlt, line 13: Did you add `conda-forge` for `rioxarray`? If so, I'd probably suggest inlining it with `- conda-forge::rioxarary` instead of adding it globally. I've seen `channel::package` being recommended in some places, I'm not entirely sure how it works, e.g. does it pull all the dependencies of `rioxarray` from conda-forge or can it pull them from the global channels too?

Well but to be honest I think that instead of that I would just make `conda-forge` the unique and global channel:
- there's no reason to depend on `pyviz` I think, all the HoloViz packages are on conda-forge, just some hours after they're published on `pyviz`
- `defaults` can be quite lagging, e.g. rasterio 1.3.7 is on conda-forge vs. `1.2.10` on defaults (which is not too bad, but still behind). That doesn't matter so much for these examples (v.s. when we're running a test suite) but still, why not using the latest when we can.
- `conda-forge` works pretty well and with the spread of the libmamba solver users don't have too wait so much when solving complex environments

On that last point, I'm actually thinking about something else where @jlstevens could shed some light on. IIRC `anaconda-project` does solve again based on the list of dependencies pinned in the lock file, instead of installing them all directly. Could it be that when `conda-forge` is involved, installing a locked project can take a very long time when you're not using the libmamba solver? In which case @sandhujasmine forget about what I said on `conda-forge`, as I think we expect some of the projects on examples to be re-used by AE5 users who can't switch easily to libmamba.

landsat/landsat.ipynb
```diff
@@ -7,7 +7,7 @@
     "# Landsat\n",
     "Written by Brendan Collins<br>\n",
     "Created: June 20, 2016<br>\n",
-    "Last updated: May 25, 2021"
+    "Last updated: May 29, 2023"
```
maximlt, line 10: As you can see on the [dev website](https://pyviz-dev.github.io/examples/landsat/landsat.html) a header is now automatically generated, so this header is no longer required and can be removed from the notebook.

landsat/anaconda-project.yml
sandhujasmine: I was going to ask if pinning each package in anaconda-project.yml is really need3ed since the anaconda-project-lock.yml already pins all packages; though that's not easy to read. I'll pin these though it would be nice if this could be automated somehow

landsat/landsat.ipynb
sandhujasmine: I see Brendan's github ID ([brendancol](https://github.com/brendancol)) - I'll add this to anaconda-project.yml

landsat/anaconda-project.yml
jbednar: I've previously argued that you should put >= pins for any package whose API you depend on, filling in the version number of whatever you wrote it for. I.e. if you used pandas 1.2 when you wrote the notebook, pin to >=1.2 in anaconda-project.yml, since you may have used syntax only available in 1.2 or later. If you tested with pandas 1.1, then put >=1.1.  That way you have a recording of what you originally did, which can help explain things. E.g. if you're wondering why a particular notebook is using some weird old way of doing things, you can see that it was written for holoviews 1.2 and can realize "oh! they just didn't know any better then! I can go ahead and update that now". 

In practice, though, as @maximlt says above, most people don't do this, so I'm not sure if there's much point if it's not enforced and not done much in practice.

landsat/anaconda-project.yml
jbednar: Note that anaconda-project is completely misleading about the :: syntax. It does *not*, as you would hope and expect, bring in just that package, or even bring in just that package and its dependencies. Last I checked, it simply added the channel listed onto the channel search list, which to me does not seem like reasonable behavior.  That's another reason to simply use c-f if you do need anything from c-f.

landsat/landsat.ipynb
jbednar: I totally don't remember who started it originally, which was presumably Brendan, but I definitely did contribute substantially to it, e.g. making the hyperspectral plots and the lookup table involved, so I should be listed as a second author.

sandhujasmine: I already saw you as the first author in the anaconda-project.yml and just added his name to it as well. I assumed that was the correct thing to do here :)

jbednar: Sure. Github blame might resolve that, but who knows.

landsat/anaconda-project.yml
sandhujasmine: **New learning:**
I understand. We don't specifically care about the `jupyter_client` dependency.
There should be some way to lock `notebook <x.x.x` and/or `python <x.x.x` such that it ensures the correct version of jupyter_client less than 8 is pulled. This is an interesting problem since we really want to know if there is a combination of package versions defined in `anaconda-project.yml` which pulls in jupyter_client <8.0. Currently this happens manually, trial and error AFAIK or using `conda search --info` but a quick test shows that it has a lower bound but not necessarily an upper bound. 🤔 

**Another followup from original question:**
I just wrote a simple script to get the min version out of the `anaconda-project-lock.yml` for any unpinned packages. This wasn't hard to do but unsure how robust it is - it just **finds the min version from all platforms and uses that as lower bound for each package that is not explicitly pinned**.  Then regenerate the lock file and test to ensure things still work on the 1 platform you're using. 

I think this should be obtained from the environment or from the lock file since presumably the env is a reflection of the lock file. Nicer if it is scripted, plus less error prone. But is the above claim in **bold** always correct? We can certainly (re)-test. Tools should make our life easier, so definitely not advocating to creating tools that become complex and must then be maintained.

The requirement of adding the bound in the `anaconda-project.yml` makes it easy to see min version of packages without going to the lock file which is nice for sure. I think we can automate this assuming that retesting after generating the lower bound results in the notebook working as before (99% of the time).

I can move these out as separate issues/topics to discuss since they apply more generally to all examples.

landsat/anaconda-project.yml
sandhujasmine: > Note that anaconda-project is completely misleading about the :: syntax. It does not, as you would hope and

Indeed - I'm not sure I understand channel priority or how it is used to solve for the packages; probably not easy to state. I had to limit it to c-f. I'm not sure how to interpret / understand :: syntax. I found few references to this syntax - saw this on [conda-managing-channels](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html):

"Essentially, the order goes: channelA::numpy-1.13_1 > channelA::numpy-1.12.1_1 > channelA::numpy-1.12.1_0 > channelB::numpy-1.13_1" 

Will ask folks who know more about conda.

sandhujasmine:
Not sure why the libmamba solver is looking for OSX packages on linux: https://github.com/holoviz-topics/examples/actions/runs/5148674083/jobs/9270715411?pr=273

@maximlt - any ideas? I assume this will not work on our public facing docs page since linux env is not getting created, is that correct?

landsat/anaconda-project.yml
```diff
@@ -10,38 +11,30 @@ examples_config:
...
+- geoviews >=1.10.0
+- holoviews >=1.16.0
+- numpy >=1.24.3
+- pandas >=2.0.2
+- rasterio >=1.2.10
+- xarray >=2023.5.0
+- numba >=0.57.0
+- param >=1.13.0
+- rioxarray >=0.14.1
+- jupyter_client =7.4.9
```
sandhujasmine, line 32: These were generated by finding the minimum version from the lock file. I don't know if that's the best process but the lock file should still work for creating the environments.

landsat/anaconda-project.yml
sandhujasmine: Resolving for now.

landsat/anaconda-project.yml
jbednar: If it worked on defaults before, can you record here why conda-forge is needed now?  Presumably everything involved must be available on defaults, just not in compatible versions?

jbednar: And can you remind us why it needs these pins? I think it was something to do with cross-platform compatibility? Usually we'd put pins like that just to indicate the minimum version needed when the notebook was written, but this is a notebook from 2016 and so presumably none of those versions are truly minimal.

jbednar (commented):
Looks good! Happy to have it merged after those updates.

landsat/anaconda-project.yml
maximlt: > And can you remind us why it needs these pins? 

Isn't it just because we asked for it in a recent discussion here? As even if that's not required, that makes maintaining the project easier on the long term.

> the minimum version needed when the notebook was written

That might just be lost? I'm happy with updated pins.

landsat/anaconda-project.yml
```diff
@@ -2,6 +2,7 @@
 name: landsat
 description: Datashading LandSat8 raster satellite imagery
 examples_config:
+  created: 2016-06-20
   maintainers:
   - jbednar
```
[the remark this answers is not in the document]
maximlt: There's no difference between authors and maintainers. IMHO that would be a little too detailed, and some work to find trace back the list of original authors and distinguish them with the maintainers... If that needs to be addressed, I'd suggest renaming it `contributors`, so that it sort of encompasses both?

landsat/landsat.ipynb
```diff
@@ -293,9 +291,22 @@
...
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
    "name": "python",
-   "pygments_lexer": "ipython3"
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.3"
```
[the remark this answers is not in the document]
sandhujasmine: Fixed

landsat/anaconda-project.yml
sandhujasmine: > Isn't it just because we asked for it in a recent discussion here? As even if that's not required, that makes maintaining the project easier on the long term.

Yes, that was my understanding. But in light of comments here: https://github.com/holoviz-topics/examples/issues/275 - I'm going to remove all the pins except the one for `jupyter_client`

landsat/anaconda-project.yml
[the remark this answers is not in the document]
sandhujasmine: I assume these tags are parsed and displayed when we render the notebooks on the https://examples.pyviz.org/ page. Is `contributors` or `authors` parsed and displayed correctly? 

Looking here, I only see maintainers: https://github.com/holoviz-topics/examples/blob/main/dodo.py#L1019-L1023
It looks like I should also add a `last_updated` tag - it certainly applies.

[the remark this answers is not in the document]
maximlt: > contributors or authors parsed and displayed correctly

Nope, it's only `maintainers` and it is a required field. Nothing set in concrete though, but it needs some changes.

>  I should also add a `last_updated` tag

There are two options for `last_updated`:
- set it in the anaconda-project.yml file
- don't set it, it's computed automatically based on the last commit date that touched the project (anything in the project), see https://github.com/holoviz-topics/examples/blob/833126e41ac1db873209b9db1e97f0ed3408385c/doc/conf.py#L136

I think I prefer the automatic approach, I don't have much confidence in people updating that field 🙃  Myself first!

landsat/anaconda-project.yml
jbednar: Note that this behavior is controlled by anaconda-project, not conda. conda does behave properly with channel specifications like that, but anaconda-project (last I looked) simply strips off the channel specifier, adds it to the list of allowed channels, and only then gets conda involved. So don't look at what conda does for ::, because that's not invoked here.

landsat/anaconda-project.yml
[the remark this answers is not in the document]
sandhujasmine: I missed this - will remove it. Thanks!

maximlt:
@sandhujasmine just a heads up, as you can see, I'm trying to debug something 🙃 Nothing to do with your updates! It seems like the dev website doesn't get properly updated, the landsat page built doesn't have any of the updates you made https://holoviz-dev.github.io/examples/landsat/landsat.html.

The docs build generate and upload an artifact that contains the whole built site, I set this up a while ago for debugging purposes. Also because we only have one dev site but multiple competing PRs can update it at the same time, so it's a fallback for users to see what the built site looks like with their changes only. The artifact can be downloaded from the summary page of all the jobs that run: https://github.com/holoviz-topics/examples/actions/runs/5220200442

I can see that this artifact also contains the older site, so I need to figure out what's going on when the evaluated branch and the temporary evaluated branch are synced during the build.

PR #274, opened 2023-06-05
[compliance] landuse_classification

sandhujasmine:
- updated lock file and removed pins for now from anaconda-project.yml
- updated catalog.yml so data loads as before
- small updates to calls since so warnings are fixed (API likely changed w/ updated packages)


landuse_classification/Image_Classification.ipynb
```diff
@@ -63,7 +62,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "scrolled": true
```
jbednar, line 66: These details should get cleared by the kernel stripping script.

landuse_classification/anaconda-project.yml
```diff
@@ -4,6 +4,7 @@ description: Image classification using the UC Merced Land Use Dataset
 examples_config:
   maintainers:
   - jbednar
+  - philippjfr
```
jbednar, line 7: Philipp did way more on this than I did:
```suggestion
  - philippjfr
  - jbednar
```

landuse_classification/Image_Classification.ipynb
sandhujasmine: Fixed!

maximlt (approved):
We did it @sandhujasmine !!! Thanks for your work, that helped me a lot to complete this PR.

--

This is not a modernization kind of PR, it's just to make this project valid and allow it to be tested and built on the CI. Talking about the build, fitting the model takes an hour on the CI, it's long compared to most examples but still reasonable (10 hours wouldn't be). One issue I observed is that the build is not deterministic, during the last one the model performed really badly and I have to re-run the build, which fortunately gave some decent result (I set the seed previously but it's not enough apparently). Not a blocker for this PR to be merged.

PR #281, opened 2023-06-16
Updated icon and favicon and text from pyviz to holoviz

jbednar:
Update contents of site to reflect name change from examples.pyviz.org to examples.holoviz.org


jlstevens:
Looks good to me!

maximlt:
I will have to update the logo. It should be the same as the one used on holoviz.org.

This PR:
image

holoviz.org:
image

maximlt (approved):

maximlt:
There would be more pyviz links to update, but that's low priority now the redirect link is set up and that would anyway be pretty annoying to do in this PR as it'd mean updating the affected projects to conform with the new linters and test/build set up. But that's definitely something worth doing later on.

PR #285, opened 2023-06-18
infra: add a pre-commit hook to maintain a top-level list of deployments

maximlt:
@jlstevens what do you think about that? It's a simple pre-commit hook that will keep updated a top-level YAML file with a list of all the deployments declared across the project files, to be consumed by some monitoring process.

I would not merge that until we think monitoring through anyway but I wanted to have a quick feedback on this from you :) Marking as draft.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+    - repo: local
+      hooks:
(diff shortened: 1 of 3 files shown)


maximlt:
Superseded by https://github.com/holoviz-topics/examples/pull/345

PR #286, opened 2023-07-28
Add t_sne_roots project

jlstevens:
Adds an interactive dimensionality reduction dashboard example with instant inspection hover.

UMAP_hover_dashboard

Wondering if we should leave the training data on hugging face or grab our own copy...


t_sne_roots/t_sne_roots.ipynb
```diff
@@ -0,0 +1,156 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b3f9e6b7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "points = hv.Points(df, ['x','y'], ['language'])\n",
+    "op = datashade(points, color_key=cc.glasbey_light, \n",
```
maximlt, line 79:
```suggestion
    "op = datashade(points, color_key=cc.glasbey_light,\n",
```

droumis:
it would be amazing if the hover tooltip is hidden or suppressed when the cursor is not near a real data point; that would make this inspection much more like what users experience for non-image plots.

I'm guessing this would require a custom hovertool and customjs callback that sets the visibility of the tooltip based on the data value

maximlt:
Oh I forgot about that bot, happy that it worked: https://holoviz-dev.github.io/examples/t_sne_roots/t_sne_roots.html

jlstevens:
@droumis That is an excellent observation! Ideally this is something we could push to the Bokeh level...

MarcSkovMadsen:
Would it be an idea to make the plot responsive? (Yes 😉)

jbednar:
> Would it be an idea to make the plot responsive? 

Yes 😉. Jean-Luc indicated that the tricky aspect of that was, well the aspect, in that the plot gets squished. But I argue that the plot should fill the template fully, but with black on either side initially and showing the complete data when zooming in. Using only the largest square plot is fine for a static plot, but as soon as one zooms in, it feels very constraining to be locked into that tiny box.

jbednar:
> Wondering if we should leave the training data on hugging face or grab our own copy...

Definitely grab our own copy.

t_sne_roots/t_sne_roots.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "import pandas as pd\n",
+    "import datashader as ds\n",
+    "import colorcet as cc\n",
+    "\n",
+    "import holoviews as hv\n",
+    "from holoviews.operation.datashader import datashade, rasterize, dynspread\n",
+    "import panel as pn\n",
+    "\n",
+    "pn.extension()\n",
+    "hv.extension('bokeh')"
```
jbednar, line 27: Are both extensions needed? I keep being confused about that, but thought only one was needed.

t_sne_roots/t_sne_roots.ipynb
```diff
@@ -0,0 +1,156 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b3f9e6b7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "points = hv.Points(df, ['x','y'], ['language'])\n",
+    "op = datashade(points, color_key=cc.glasbey_light,\n",
```
jbednar, line 79: How many distinct languages are there, and are they sorted by popularity? At a glance it looks like 20 or so main categories, and if so Category20 might give more vibrant colors, or maybe glasbey_category10 if there are 20 main ones but then lots of rare categories.

jbednar (approved):
Looks good, thanks! Looking forward to seeing that second pass eliminated. Can you maybe also open an issue that will stay open once this is merged, asking you to update this project once a given task list is addressed, where each item in the task list is an open Bokeh or HoloViews issue that needs to be fixed and released before this example can be updated with one pass?  If new issues are needed on either of those repos, please do open those as well.

PR #337, opened 2023-12-06
Updated versions of t_sne_roots notebooks

jlstevens:
Updates and supercedes https://github.com/holoviz-topics/examples/pull/286

In this updated version, the default dashboard uses client-side color mixing (i.e. using `rasterize`). The previous datashaded version is included both as a point of comparison and because it supports `dynspread` (which `ImageStack` currently does not).


jlstevens:
Addressing an earlier question by @jbednar 

> How many distinct languages are there, and are they sorted by popularity? At a glance it looks like 20 or so main categories, and if so Category20 might give more vibrant colors, or maybe glasbey_category10 if there are 20 main ones but then lots of rare categories. Or Category20 + glasbey_light, in that latter case.

There are 47 languages and they are not sorted.  Here are the colormap options I've tried:

`cc.glasbey_light` (my original choice)
image

`cc.b_glasbey_category10`
(I found category 10 but not 20 unless you meant `cc.b_glasbey_bw_minc_20` which is next)

image

`cc.b_glasbey_bw_minc_20`
image

`cc.b_glasbey_category10 + cc.glasbey_light`

image

Can't say I have a strong opinion between these!

jbednar:
Category20 is from Bokeh: https://docs.bokeh.org/en/latest/docs/reference/palettes.html#d3-palettes

Can you use df.cat.value_counts() to get a list of the categories by popularity, then use Category10 or Category20 for those top 10 or 20 categories, then glasbey_light for the rest?

jbednar:
Putting them side by side (glasbey_light, category20, category10) shows the colors do get more vibrant when the more common categories use the brighter category10 colors, and I can see more distinct categories in the figure (looking outside the big are that's orange on the right):

image

If you use `glasbey_category10` (already concatenated in colorcet) does it still need both hard-to-explain lines? Putting the most frequent categories with the most intense colors seems like a reasonable thing to record how to do, if it's not too crazy as code.

jlstevens:
With https://github.com/holoviz/holoviews/pull/6024 dynspread now works with `ImageStack`:

t_sne_imagestack

However, it is a fair bit slower than the datashaded version:

t_sne_datashaded

jbednar:
Can you figure out why that would be?

droumis:
For cmap comparison, here is the [image Christopher produced](https://huggingface.co/datasets/cakiki/roots-tsne-data#plotting) on the far left and then the three most recently being considered in this PR:
image

Azaya89:
Hi @jlstevens, what's the status of this PR?

droumis:
@Azaya89 , I had intended to continue the work in this PR but I haven't had time yet

PR #343, opened 2024-01-16
infra: no longer use pyviz-developers

maximlt:
Let's see how it goes


maximlt:
Merging. The docs workflow might break when it runs on merged PRs, let's see.

PR #357, opened 2024-02-02
gull_tracking: modernized project example

Azaya89:
- [x] Modernized example
- [x] Added explanations for the code
- [x] Discuss hosting the file on our S3 due to the lack of the *content-length* in the response from Zenodo preventing any progress bar to be displayed by anaconda-project?


gull_tracking/gull_tracking.ipynb
```diff
@@ -37,13 +35,8 @@
...
    "outputs": [],
    "source": [
-    "hv.extension('bokeh')\n",
-    "\n",
-    "map_tiles  = EsriImagery().opts(alpha=1.0, width=800, height=800, bgcolor='black')\n",
-    "points     = hv.Points(df, ['longitude', 'latitude'])\n",
-    "rasterized = shade(rasterize(points, x_sampling=1, y_sampling=1, width=800, height=800), cmap=fire)\n",
-    "\n",
-    "map_tiles * rasterized"
+    "df.hvplot.points(x='longitude', y='latitude', rasterize=True, tiles='EsriImagery',\n",
```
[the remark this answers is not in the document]
Azaya89: Yeah, I figured it out. The difference was in setting the `cnorm` parameter to 'eq_hist'. It wasn't set to that before.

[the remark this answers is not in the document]
jbednar: Yep, `eq_hist` is pretty important! It's not the default in Bokeh because that would be surprising for most data, but for Datashader output it's generally the best initial approach, switching to another simpler type of normalization only once you've seen what the data looks like and are ready to focus on a well-behaved subset of it.

gull_tracking/gull_tracking.ipynb
```diff
@@ -37,13 +31,12 @@
...
    "outputs": [],
    "source": [
-    "hv.extension('bokeh')\n",
-    "\n",
-    "map_tiles  = EsriImagery().opts(alpha=1.0, width=800, height=800, bgcolor='black')\n",
-    "points     = hv.Points(df, ['longitude', 'latitude'])\n",
-    "rasterized = shade(rasterize(points, x_sampling=1, y_sampling=1, width=800, height=800), cmap=fire)\n",
+    "df = pd.read_csv('./data/HG_OOSTENDE-gps-2018.csv', usecols=['location-long', 'location-lat'])\n",
+    "df.columns = ['longitude', 'latitude']\n",
+    "df.loc[:,'longitude'], df.loc[:,'latitude'] = en_ll(df.longitude, df.latitude)\n",
```
jbednar, line 36: Please add some text to explain this step, similar to the following:

The CSV file contains latitude and longitude information in degrees. Here we will want to overlay this data onto geographic mapping tiles that use the [Web Mercator projection](https://en.wikipedia.org/wiki/Web_Mercator_projection), so we use a utility from HoloViews to project the spherical coordinates into the tiles' coordinate system to allow them to be overlaid.

gull_tracking/gull_tracking.ipynb
```diff
@@ -37,13 +31,12 @@
...
     "\n",
-    "map_tiles * rasterized"
+    "df.hvplot.points(x='longitude', y='latitude', rasterize=True, tiles='EsriImagery',\n",
+    "                 cmap='fire', cnorm='eq_hist', width=800, height=800)"
```
jbednar, line 39: I think the call should include `dynspread=True` so that dots are visible when you zoom in.

jbednar (approved):
Nice and clean, thanks!

gull_tracking/gull_tracking.ipynb
jbednar: You might also consider removing the `df.columns = ['longitude', 'latitude']` step and simply keeping the original names; I don't think it's any clearer to rename them and it adds an extra line to the file.

maximlt (changes requested):
image

I feel it would be good to describe more the dataset before displaying it. I'd suggest showing the output of `df.head()` and `len(df)`. The output of `df.describe()` wouldn't be very interesting, I think, as we only read the lat/lon columns (the CSV has more columns). The text says there are 2.4 million points in the dataset but as someone looking at this example I'd rather see the output of a code cell show that :)

image

I was a bit stumped by the fact that the map wasn't centered on the dataset. Running the notebook, I found out that there are 4 points farther West and North.

image
image

(These data points are actually displayed on the map, but they're very small and have a dark color. We use `dynspread=True` to deal with this kind of case, maybe its settings like `threshold` and `max_px` need to be adjusted to better display these isolated points. @jbednar)

I think the example would look better if there was a plot focused on where most of the points are. For example, after removing these points  (e.g. `df = df[(df['location-long'] > -14) & (df['location-lat'] < 53.3)]`), the map looks better I think:

image

maximlt:
I stored a copy of the file on our S3 and updated the `anaconda-project.yml` accordingly.

maximlt (approved):
LGTM! Thanks for the latest changes, merging :)

PR #359, opened 2024-02-08
Updates the ship_traffic notebook to match AE5 deployment

jlstevens:
Updates the project to match the AE5 version. 

Fixes the legends and adds instant inspection although this approach is somewhat ugly.

image


droumis (changes requested):
I think the following can be removed, no? I don't see where it is used

```python
def convert(x):
    try:
        return int(x)
    except:
        return 0
```

droumis:
Is there a workaround or alternative for getting the images from marinetraffic website? Looks like they are now blocking programmatic access now

droumis:
I removed the unnecessary code.
I will file an issue about the restricted access to the image data.

will merge as-is now

droumis (approved):

PR #362, opened 2024-02-27
[DO NOT MERGE] Isaiah's Dummy example

Azaya89:
I created a dummy example in order to verify all the steps required in creating a new example project from scratch, per https://examples.holoviz.org/contributing.html#add-a-project

I think the only thing we need to modify in the instructions is explicitly mentioning that the `doit` commands should be run from the top level of the repository. Otherwise, everything else is clear and straight-forward.


maximlt (changes requested):
I'm impressed it went so smoothly :)

Could you update this example to include a deployable panel app? I.e:
- the notebook must contain something like `<app>.servable()`
- the project YAML must be updated with a new `dashboard` command and that command must be specified in `examples_config.deployments`

droumis:
Nice work! Do you want to add the suggestion for the instructions to your contributing guide PR?

Azaya89:
> Nice work! Do you want to add the suggestion for the instructions to your contributing guide PR?

Yes, I will do that tomorrow.

maximlt:
You did great creating this dummy example! I'm going to close this PR now (we can still re-open it later if need be).

PR #377, opened 2024-04-08
infra: add a step to validate the content of notebooks

maximlt:
Precisely, to warn/fail when a notebook contains code outputs when it shouldn't, or doesn't contain any when it should (`skip_notebooks_evaluation: True`). The goal of this additional validation step is to make it less likely to commit a notebook with code outputs, which is a common gotcha.

cc @jtao1 @Azaya89

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -464,6 +464,29 @@ def proj_env_vars(project, filename='anaconda-project.yml'):
     return env_vars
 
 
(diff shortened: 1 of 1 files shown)


Azaya89:
Is there a `doit` command we should run to activate this step?

maximlt:
It should run by default when you call `doit validate:<project>`. It will also run on the CI.

Issue #374, opened 2024-03-27
Last updated date does not match GitHub last commit date


ahuang11:
Based on https://examples.holoviz.org/gallery/gapminders/gapminders.html, I saw that it was updated yesterday! 
<img width="844" alt="image" src="https://github.com/holoviz-topics/examples/assets/15331990/56e3460b-75e9-43e7-ac29-a76e2ae213d7">

But looking at the GitHub repo, I saw it was updated 5 months ago!
<img width="1281" alt="image" src="https://github.com/holoviz-topics/examples/assets/15331990/5ecdac0c-9925-4bfb-ad2a-8a875dd5e297">

maximlt:
Fixed by https://github.com/holoviz-topics/examples/pull/425

PR #387, opened 2024-05-23
Fix last update date

hoxbro:
Fixes https://github.com/holoviz-topics/examples/issues/374

I haven't tested it out; I'm pretty sure this is why the last update date doesn't show up correctly. I think this is where `fetch-depth` should be set, but please correct me if it is not.

### DIFF
--- a/.github/workflows/build_one.yml
+++ b/.github/workflows/build_one.yml
@@ -56,6 +56,7 @@ jobs:
     timeout-minutes: 60
     steps:
     - uses: actions/checkout@v3
+      fetch-depth: 0  # To get correct last update data
     - uses: conda-incubator/setup-miniconda@v2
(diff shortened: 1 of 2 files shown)


maximlt (approved):
Good catch I think. I wonder how you find the time to look into the issues of all the repositories we have!

hoxbro:
Unfortunately, it didn't seem to work :upside_down_face: 

I don't know when I will have time to dive deeper into why this is the case.

maximlt:
> Unfortunately, it didn't seem to work 🙃
> 
> I don't know when I will have time to dive deeper into why this is the case.

Ended up with the same fix in https://github.com/holoviz-topics/examples/pull/425, not sure why it didn't work here.

PR #397, opened 2024-06-04
Validate that notebook<7 is used

maximlt:
The Workbench platform doesn't yet support deploying read-only notebooks from environments that use `notebook` v7 or greater. This PR adds a validation step that ensures that a project that declares such a deployment locks to a previous version of `notebook`.

### DIFF
--- a/.github/workflows/test_one.yml
+++ b/.github/workflows/test_one.yml
@@ -78,6 +78,8 @@ jobs:
       run: doit validate_project_file:${{ inputs.project }}
     - name: validate anaconda-project lock file
       run: doit validate_project_lock:${{ inputs.project }}
+    - name: validate notebook v7 pinned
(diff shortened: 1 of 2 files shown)


Azaya89:
I dunno if this brings us back but what Workbench platform problem is this PR avoiding? 'Cos I was having no issues locally building the docs from my updated notebooks @maximlt

maximlt:
Yes this is unrelated to building the docs, instead, it's related to deploying a read-only notebook, like for example https://boids-notebook.holoviz-demo.anaconda.com/notebooks/boids.ipynb. Making such a deployment isn't part of the test suite, it only happens *after* we merge a PR, so you don't know whether the deployment fails, unless you check these Github workflows https://github.com/holoviz-topics/examples/actions/workflows/deploy.yml.

Azaya89:
OK. Thanks for the clarification!

PR #399, opened 2024-06-24
Update ship_traffic and template resource profiles

maximlt:
ship_traffic to medium and template to default.

cc @jlstevens


jlstevens:
Looks good, thanks!

PR #409, opened 2024-08-05
Add sorting by last updated date or title

droumis:
This won't look like it's working yet on the deployed site because of https://github.com/holoviz-topics/examples/issues/374

But here it is working on my local build:

https://github.com/user-attachments/assets/c21ad992-fa24-41e3-a8ed-94206ab0e0dd


jlstevens:
Thank you!

Having the option to sort will make the tile view even more useful. Sorting by name and last updated are the most important things but if we have the metadata for it, we can also imagine sorting by author name, creation date etc.

droumis:
> Thank you!
> 
> Having the option to sort will make the tile view even more useful. Sorting by name and last updated are the most important things but if we have the metadata for it, we can also imagine sorting by author name, creation date etc.

Sure, but in future PRs

droumis:
alright to merge?

jlstevens:
Thanks!

PR #412, opened 2024-08-06
nyc_taxi patch update

### LINKED ISSUE #415: Fix nyc_taxi
The map is missing and the background color is wrong:

We previously used stamen terrain tiles, but stamen tiles are not yet available. Start by making the bgcolor white and use the map tiles in the second image below

image

use bgcolor = 'black' or use a darker map tiles option

image

The panel dashboard notebook on nyc_taxi looks good:

image

Azaya89:
I added the `Plotsize` function to the nyc_taxi geographic notebook to enhance the rendering of the images when shown on the website.

- [x] Also fixes #415


droumis:
This plot needs to be fixed:

image

droumis:
The colormap is inverted here:

image

Azaya89:
Here is how both plots look now:

plot1
plot2

maximlt (approved):

PR #414, opened 2024-08-13
Demo for video - DO NOT MERGE!

Azaya89:
This is just a demo PR for the video in our contribution website.


maximlt (commented):
Not sure this was discussed already but did you consider using the `template` project? https://github.com/holoviz-topics/examples/tree/main/template It's meant to be used as a project contributors should copy/start from. There's special handling of this project in various places, for instance to avoid adding it to the production website (it is present on the dev website).
I'd like to avoid having two projects like that.

Azaya89:
> Not sure this was discussed already but did you consider using the `template` project? https://github.com/holoviz-topics/examples/tree/main/template It's meant to be used as a project contributors should copy/start from. There's special handling of this project in various places, for instance to avoid adding it to the production website (it is present on the dev website). I'd like to avoid having two projects like that.

I created this demo PR just to show the whole process of adding a new example. Using the template example itself wouldn't show how to create a new example fully, even though the data used for this demo example was gotten from the template directory.

Nevertheless, we can still delete this PR. It wasn't meant to be merged. It is only useful in the video, and not anywhere else.

PR #420, opened 2024-08-29
Replace the full PR modernization checklist template with a link to wiki

droumis:
Since the NF SDG is done, the modernization checklist is more of an optional resource and may be confusing to show the whole thing without context


maximlt:
Agree with that. Do you want to put the sentence you added as a Markdown comment (`<!-- foo -->`)? I don't have any strong preference.

maximlt (approved):

PR #427, opened 2024-10-10
World Cup example patch

Azaya89:
I deleted some of the files in the world cup test directory as they were not used in the notebook, even though they were part of the files downloaded when the notebook is run.


maximlt:
Thanks! Could you show the size of the test data files left for this example?

Azaya89:
> Thanks! Could you show the size of the test data files left for this example?

2.3 MB  and 130 kb

maximlt:
> I deleted some of the files in the world cup test directory as they were not used in the notebook, even though they were part of the files downloaded when the notebook is run.

Ok I get that, one of the zip files downloaded via anaconda-project contains some extra JSON files which are not used by the example. It's unfortunate we can't get just the file we need but the whole zip size is about 70MB so it's not too bad.

I've made some small changes:
- Even smaller test data file sizes (picking 1% of the players, and the events of these players only).
- Removed my name from the authors
- Some formatting changes, notably making a few very long code lines shorter as they would be way too long for a good experience on mobile

Azaya89:
> I've made some small changes:
> 
> * Even smaller test data file sizes (picking 1% of the players, and the events of these players only).
> * Removed my name from the authors
> * Some formatting changes, notably making a few very long code lines shorter as they would be way too long for a good experience on mobile

Thank you!

PR #436, opened 2024-10-22
Include filter.js in the gallery index only

maximlt:
Other pages were raising this error:
image


maximlt:
@droumis for awareness

PR #438, opened 2024-10-23
Improve handling of the labels

maximlt:
- No more SVG, they're now sphinx-design badges
- Label filter button ordered by count on the gallery landing page
- Displayed in the header


maximlt:
cc @droumis for awareness

Issue #93, opened 2020-02-12
Issues with attractors_panel


jbednar:
The attractors_panel notebook works well generally, but it has a few open issues:

- [x] 1. Saving to a filename only works if that file already exists on disk.  Looks like I already created an issue for this years ago, but it's still a problem: https://github.com/holoviz/param/issues/304
- [x] 2. If the "+" button is selected (speeding up the player) or the N is increased (slowing down the player) or a slow machine is used, hitting Play can cause drawing/update events to be generated faster than they can be completed. The display can then lag behind the controls in a confusing way, and hitting Stop appears not to have an effect (though it will eventually stop once the queue of events has been consumed). It would be nice to be able configure the player to drop outstanding events when adding new ones, so that it skips ahead rather than getting a confusing backlog. 
- [x] 3. There is too much code in attractors_panel.ipynb.  In practice classes like these would be defined in a .py file, and it would be nice to make this file illustrate good practice in splitting code between notebooks and Python modules, by moving the non-GUI code into a module and showing how the notebook works well for GUI construction.  We'd need to be able to link to the Python file from the notebook where appropriate.
- [x] 4. The Player widgets look very different (and worse) in the rendered HTML page at https://examples.pyviz.org/attractors/attractors_panel.html than they do in Jupyter or in deployed notebooks; much larger and solid colors rather than rounded buttons, and overlapping the subsequent text. Problem with CSS in Sphinx or nbsite?
- [x] 5. The web page https://examples.pyviz.org/attractors/attractors_panel.html used to render the equations, but currently only shows the ASCII source for them, even though similar (?) equations work on https://examples.pyviz.org/attractors/attractors.html

jbednar:
Point 3 is addressed by https://github.com/pyviz-topics/examples/pull/94 . 

Point 5 seems to be working fine today after shift-reloading the page; we'll see if there continue to be issues. The equations still briefly render the ASCII text, but then reformat into the typeset equations.

jbednar:
To illustrate problem #4:
![image](https://user-images.githubusercontent.com/1695496/74477125-53f39f00-4e70-11ea-934a-b0afb29e44ad.png)

Azaya89:
Based on #381, would you say these bugs have been addressed? @jbednar

jbednar:
[...]

PR #457, opened 2024-11-26
Fix the strange attractors app

maximlt:
Addresses https://github.com/holoviz-topics/examples/issues/93


jbednar:
Looks good! It worked well for me at https://attractors.holoviz-demo.anaconda.com/attractors_panel with reasonable settings. I was able to get the player buttons to become unresponsive, but only after yanking the number of points way up and turning on line drawing, so I think it's good now. Thanks, @maximlt and @Azaya89 !

PR #460, opened 2024-11-29
[compliance] walker_lake

maximlt:
Main changes as I had some issues getting the example to run:

- No more intake: intake + intake-xarray was raising some errors, replaced by anaconda-project downloads
- rioxarray used to open the files
- Bigger chunks: even with bigger chunks I find the example slow with Dask, the data fits in the memory, yet it's a good example to show how to use Dask (maybe there are other knobs to tweak, maybe some performance optimisations in our HoloViz toolset)
- No more resampling/regridding with Datashader: After the package upgrade, the mean/sum plots were showing very little data (I'm not sure why), enough though to skip the regridding branch. It worked better when regridding was forced. I imagine that geo users wouldn't regrid their data with datashader+geoviews. I tried `rioxarray.project_match` which worked great but is not Dask compatible (it just loads the whole dataset in memory). `xarray.interp_like` worked perfectly and gave the same results as `rioxarray.project_match`.

Note that this is not a full update of this example (a la NF SDG), it's just to adapt it to the (not so) new infra (e.g. more validation checks) like we did a year ago for most of the examples.


walker_lake/walker_lake.ipynb
```diff
@@ -131,15 +135,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "print(\"The shape of the DataArray is :\", landsat_5_img.shape)\n",
```
jlstevens, line 134: Why were these print statements removed? The idea is to give the user some information about what is loaded in a friendly way...

walker_lake/walker_lake.ipynb
```diff
@@ -131,15 +135,15 @@
...
-    "print(\"With attributes:\\n \", '\\n  '.join('%s=%s'%(k,v) for k,v in landsat_5_img.attrs.items()))"
+    "assert landsat_5_img.rio.crs == landsat_8_img.rio.crs\n",
+    "print(landsat_5_img.rio.crs)"
```
jlstevens, line 139: There should be a friendly message if you think the `crs` object is to be shown here. Looks like it might be a stray debug print to me right now...

walker_lake/walker_lake.ipynb
```diff
@@ -148,7 +152,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "crs=ccrs.UTM(11)"
+    "crs = ccrs.epsg(landsat_5_img.rio.crs.to_epsg())"
```
jlstevens, line 155: This looks better than the hardcoded `UTM(11)` as long as it comes out as the same thing!

walker_lake/walker_lake.ipynb
```diff
@@ -157,7 +161,16 @@
    "source": [
     "## Computing the NDVI (1988)\n",
     "\n",
-    "Now let us compute the [NDVI](https://en.wikipedia.org/wiki/Normalized_difference_vegetation_index) for the 1988 image. Note that we need to promote the ``DataArray`` format as returned by ``rasterio`` to an xarray ``DataSet``. This restriction should be lifted in future (see geoviews issue [209](https://github.com/pyviz/geoviews/issues/209))."
```
jlstevens, line 160: Glad to see this restriction go away!

walker_lake/walker_lake.ipynb
```diff
@@ -187,10 +197,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "landsat_8_img.data[landsat_8_img.data==-9999] = np.NaN  # Replace the -9999\n",
-    "ndvi8_array = (landsat_8_img[4]-landsat_8_img[3])/(landsat_8_img[4]+landsat_8_img[3])\n",
-    "ndvi8 = ndvi8_array.to_dataset(name='ndvi')[['x','y', 'ndvi']]\n",
-    "client.persist(ndvi8)"
+    "ndvi8 = (landsat_8_img[4] - landsat_8_img[3])/(landsat_8_img[4] + landsat_8_img[3])\n",
+    "# ndvi8 = ndvi8_array.to_dataset(name='ndvi')[['x','y', 'ndvi']]\n",
```
jlstevens, line 201: Stray comments?

walker_lake/walker_lake.ipynb
```diff
@@ -200,25 +209,21 @@
...
-    "y_range = (int(max(ndvi5.y.min(), ndvi8.y.min())), int(min(ndvi5.y.max(), ndvi8.y.max())))\n",
-    "\n",
-    "plot_width = min(ndvi8.dims['x'], ndvi5.dims['x'])\n",
-    "plot_height = min(ndvi8.dims['y'], ndvi5.dims['y'])\n",
-    "cvs = ds.Canvas(plot_width=plot_width, plot_height=plot_height,\n",
-    "                x_range = x_range, y_range = y_range)\n",
-    "\n",
-    "ndvi5 = cvs.raster(ndvi5, agg= rd.mean('ndvi'))\n",
-    "ndvi8 = cvs.raster(ndvi8, agg= rd.mean('ndvi'))"
+    "ndvi8 = ndvi8.interp_like(ndvi5, method=\"nearest\")\n",
```
jlstevens, line 225: Very nice! I am assuming the result is very close to the old method.

jlstevens:
I haven't run the notebook but I have reviewed the diffs. Overall, the changes look like a good improvement! Some of the old hacks and limitations are gone and the approaches used in this PR are good examples of modern, best practice.

The main tradeoff I see here is the new function to glob the files and the verbose, explicit list of downloads in the `anaconda-project.yml`.  However, this is probably better at this point than having to introduce intake as a dependency - what I would like to see is a more compact approach that doesn't depend on intake (e.g. maybe `conda-project` downloads could support globbing?)

There are some minor comments to address but otherwise I am happy to merge.

walker_lake/walker_lake.ipynb
maximlt: I removed it because the xarray repr displays that information nicely too. Maybe it wasn't the case when the original example was created? I can add it back it you prefer!

image

walker_lake/walker_lake.ipynb
maximlt: Yes it does :) It took me a little while to find the right incantation. Ideally, we'd just accept `crs = landsat_5_img.rio.crs`.

walker_lake/walker_lake.ipynb
maximlt: Happy too! I tried and it just worked!

walker_lake/walker_lake.ipynb
maximlt: Removed in https://github.com/holoviz-topics/examples/pull/460/commits/d2fda252708eb46df3a31c315d805a5d92d5fbbd

walker_lake/walker_lake.ipynb
maximlt: Yes, it is. I also compared it with the output of `rioxarray.project_match` and there was no difference.

maximlt:
@jlstevens in https://github.com/holoviz-topics/examples/pull/460/commits/e7a18f425eafe27d43da134c9a3c033abfed074d I removed from the `downloads` section the bands not needed for the notebook to run. I did so as running `anaconda-project prepare` too a pretty long time (downloads aren't parallel apparently). With this change, we're a bit closer in spirit to the previous version I think, where Intake/Dask was used to only fetch the bands required for the computation. Let me know what you think.

jlstevens:
@maximlt Only downloading the files we are using in the notebook makes a lot of sense to me!

The only thing I would do about this is to write a comment in the `anaconda-project.yml` briefly mentioning the other files (i.e. bands) available on S3 so we don't forget they are there (maybe an interested party will find an interesting use for this data).

walker_lake/walker_lake.ipynb
jlstevens: Yes, I don't think it had such a fancy repr back then!

PR #486, opened 2024-12-17
Revise Volume Imaging Ex for use of newly released `panel-neuroglancer`

droumis:
The primary approaches have been reworked around the installable package [panel-neuroglancer](https://github.com/panel-extensions/panel-neuroglancer/), rather than depending on a in-notebook coded implementation.

The in-notebook coded implementation was retained but reframed to be about understanding the underlying implementation and about further customization, with added some explanation of the machinery.

CustomNeuroglancer

---

⚠️  It's already known that the in-notebook apps are not rendering properly on the docs site or deployment, and this should NOT block review approval. 

image

image


droumis:
I was trigger-happy and merged this PR before the last commit was finished running through the CI because it only contained minor wording style changes based on the review feedback, and since the CI was passing prior to that commit. 

But…. I think that messed up the github workflows chain of dependencies which maybe expects that the PR CI finished before merging because now the deployed main website does not contain the wording updates made in the last commit of the PR and the test+build+deploy action is failing. 

IMO, it’s not urgent to address, but I wanted to document that deployed site does not match the latest (squashed) commit that has now been merged into main. I’m tempted to click ‘Revert’ on this merged PR page and see if that allows the PR CI to rerun in full successful and then I can remerge the PR and wipe everything under the rug, but I’m worried that will break things on main/deployed website somehow and with some of our more infra-familiar team away for holidays, I have a very low risk tolerance so I think things will stay as they are for a few more days.

lesson learned

maximlt:
> which maybe expects that the PR CI finished before merging

Yep (I should put together a diagram of the infra here). When a PR is opened and its notebooks are evaluated, a temporary branch is created to which the notebooks are pushed. When the PR is merged, the evaluated notebooks are fetched from that temporary branch and pushed to the `evaluated` branch; before building the site all the evaluated notebooks of all the examples are collected from `evaluated`.

I had set up the GH repo to only allow merging if the `setup` part of the CI passes.
image

I've done that a long time ago and am not sure why I didn't pick instead the last step that builds the website. I'll update it to pick this one, we'll quickly see if there was a good reason not to set it up this way!

PR #487, opened 2024-12-19
fix: Update lock for volumetric_imaging

hoxbro:
Annoyed by the warning in Discord.

image

Looking at the [logs](https://github.com/holoviz-topics/examples/actions/runs/12408415030/job/34640087244). I'm pretty sure it is related to a bad lock file related to pip. I have updated it here.  

image


droumis:
thank you!!

PR #488, opened 2025-01-02
Volumetric Imaging: dummy edit to trigger a new build

maximlt:
See https://github.com/holoviz-topics/examples/pull/486#issuecomment-2551300387


maximlt:
@droumis hmm requiring the _Documentation_ status check sort of blocks merging a PR so that's not the right way.

image

We should perhaps look into defining *deployments*.

image

maximlt:
Ah, the dev site includes again the broken neuroglancer views, is this intentional?

image

droumis:
> Ah, the dev site includes again the broken neuroglancer views, is this intentional?

yes

PR #493, opened 2025-01-30
Add standalone app to volumetric imaging example

droumis:
This is a reversal of my previous position that a standalone app should not be included for the volumetric imaging example. Due to complications on the technical deployment side of things with a notebook that doesn't have a `.servable`, and because of the argument that part of the notebook includes using the `panel neuroglancer` view as a component in a larger dashboard, I've come around to thinking that it's fine to now include a standalone app example.


maximlt (approved):
Let's see if the neuroglancer viewer gets displayed on the deployment or not.

droumis:
strange CI error...

PR #495, opened 2025-02-03
Remove standalone deployment for volumetric app

droumis:
Remove standalone deployment, see https://github.com/holoviz-topics/examples/pull/493

### DIFF
--- a/volumetric_imaging/anaconda-project.yml
+++ b/volumetric_imaging/anaconda-project.yml
@@ -11,7 +11,7 @@ examples_config:
   categories:
   - Neuroscience
   title: "Volumetric Imaging"
-  deployments:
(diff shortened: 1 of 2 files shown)


volumetric_imaging/anaconda-project.yml
```diff
@@ -13,7 +13,7 @@ examples_config:
   title: "Volumetric Imaging"
   deployments:
```
maximlt, line 14:
```suggestion
  # deployments:
```

maximlt:
The CI doesn't handle removing automatically the deployment (I think it only handles removing the deployments of a project that has entirely been removed). I manually brought the deployment down and removed the project on Workbench.

PR #509, opened 2026-06-16
feat: Convert anaconda-project files to pixi format

hoxbro:
Making a script which can convert anaconda-project + lock to pixi format. The script tries to be as true as possible to the existing lock format. For the anaconda-project.yaml I have chosen not to add the deployment section.

Have locally (linux) run:

``` bash
for i in $(fd 'pixi.toml' -x dirname); do
    cd $i
    pixi install
    if $(pixi task list | grep -q download); then
        pixi run download
    fi
    cd ..
done
```

Keeping anaconda-project files for now.

Used Opus for the script.


convert-pixi.py
```diff
@@ -0,0 +1,929 @@
...
+    names: list[str] = []
+    single = len(downloads) == 1
+    for key, spec in downloads.items():
+        task = "download" if single else f"download-{key}"
+        names.append(task)
+        url, filename = spec["url"], spec["filename"]
+        dest_dir = posixpath.dirname(filename) or "."
+        if spec.get("unzip"):
+            archive = f"{dest_dir}/.{key}.zip"
+            cmd = f"mkdir -p {dest_dir} && curl -fsSL -o {archive} {url} && unzip -o {archive} -d {dest_dir} && rm {archive}"
```
hoxbro: We should account for how the nested folder.

datashader_dashboard/pixi.toml
```diff
@@ -0,0 +1,45 @@
...
+
+[target.osx-64.dependencies]
+libgfortran = "*"
+
+[target.osx-arm64.dependencies]
+libgfortran = "*"
+
+[tasks]
+dashboard = "panel serve --rest-session-info --session-history -1 datashader_dashboard.ipynb --show"
+census = "anaconda-project prepare --directory ../census\nDS_DATASET=census panel serve datashader_dashboard.ipynb"
```
hoxbro: not optimal.

hoxbro: Manually changed in [989438f](https://github.com/holoviz-topics/examples/pull/509/commits/989438f993bcecb40f1ce05e7c3b88036c3e5e6a)

attractors/pixi.toml
```diff
@@ -0,0 +1,41 @@
+[workspace]
+name = "attractors"
+channels = ["https://repo.anaconda.com/pkgs/main", "https://repo.anaconda.com/pkgs/msys2"]
```
maximlt: Would `main` and `msys2` suffice?

attractors/pixi.toml
```diff
@@ -0,0 +1,41 @@
...
+platforms = ["linux-64", "osx-64", "win-64", "osx-arm64"]
```
maximlt, line 4: If it's possible, I wouldn't mind dropping osx-64 part of this conversion.

attractors/pixi.toml
```diff
@@ -0,0 +1,41 @@
...
+numba = ">=0.59.1"
+datashader = ">=0.16.0"
+colorcet = ">=3.1.0"
+holoviews = ">=1.18.3"
+hvplot = ">=0.9.2"
+panel = ">=1.4.1"
+param = ">=2.0.1"
+pip = "*"
+setuptools = "*"
+wheel = "*"
```
maximlt: Do you know why `wheel` is required?

bay_trimesh/pixi.toml
```diff
@@ -0,0 +1,44 @@
...
+[target.osx-arm64.dependencies]
+libgfortran = "*"
+
+[tasks.notebook]
+cmd = "jupyter notebook bay_trimesh.ipynb"
+depends-on = ["download"]
+
+[tasks.download]
+env = { FILENAME = "data" }
+cmd = "mkdir -p $FILENAME && curl -fsSL -o .DATA.zip https://s3.amazonaws.com/datashader-data/Chesapeake_and_Delaware_Bays.zip && python -m zipfile -e .DATA.zip $FILENAME && rm .DATA.zip"
```
maximlt, line 43: Does what work on Windows? Or do users have to be in WSL2 or in a special shell?

carbon_flux/pixi.toml
```diff
@@ -0,0 +1,47 @@
+[workspace]
+name = "carbon_flux"
+channels = ["nodefaults", "conda-forge"]
```
maximlt: I don't think pixi needs the `nodefaults` channel.

attractors/pixi.toml
hoxbro: Yes. The conversion script force to match all dependencies in the anaconda-project-lock file. If can find it it will add it as an explicit dependency. 

We can clean this up in a follow up PR, I would really love to not do to many manual interventions with this PR.

bay_trimesh/pixi.toml
hoxbro: Curl is supported on newer windows: https://learn.microsoft.com/en-us/windows/curl/

Though it does not support `-fsSL`. Will add it as a conda dependency in a follow up PR.

attractors/pixi.toml
hoxbro: I have been tempted, but I would rather not do it in this PR.

attractors/pixi.toml
hoxbro: I tried will give it a go again.

hoxbro: Done in [4978651](https://github.com/holoviz-topics/examples/pull/509/commits/497865151fb3391f5efbd350a24d750812e3cc44)

carbon_flux/pixi.toml
hoxbro: Done in [4978651](https://github.com/holoviz-topics/examples/pull/509/commits/497865151fb3391f5efbd350a24d750812e3cc44)

bay_trimesh/pixi.toml
hoxbro: It is built-in to git bash. 

image

attractors/pixi.toml
hoxbro: It is the same reason why `libgfortran` has been added

hoxbro: I have added a small comment to the convert script here: [d810d7c](https://github.com/holoviz-topics/examples/pull/509/commits/d810d7c2c2a3af0a1123438804b649a3b352332e)

maximlt: Ok, I understand that anaconda-project adds these dependencies internally while solving the environment. I agree with keeping them like you did.

